PR #1, opened 2016-03-15
Add ShareLatex CI

sumith1896:
This CI comes handy.
@asmeurer I have used your ID for the badge link. You'll need to enable it from here https://www.sharelatex.com/github/


sumith1896:
Thanks @asmeurer

Issue #11, opened 2016-03-17
Citation list


Upabjojr:
Let's build a list of citation.

asmeurer:
I was planning on starting a .bib file, which we can just dump citations into. Anyone should feel free to add to it (everyone who has expressed interest should have push access).

certik:
Perfect.

ashutoshsaboo:
@asmeurer +1 to that! That seems to be the best solution for this! :smile: :+1:

certik:
Looks like GitHub is slowly turning into a Facebook... :beer::beer::beer::beer::beer::beer::beer::beer::beer::beer::beer::beer::beer::beer::beer::beer::beer:

scopatz:
Oh my goodness, make it stop!

asmeurer:
That's what I thought, but everyone else seems to love it, so ¯\_(ツ)_/¯.

certik:
I like the +1 and -1, so that you can +1 a comment, without writing a comment (and thus bothering everybody). I think that's good, but they should make it much less visible.

sumith1896:
Imagine notifications for this :o that's a nightmare :smile:

asmeurer:
Well the whole point is that people were tired of being notified of +1's from random people, so they added a way to +1 without a notification.

asmeurer:
There's a bib file there now. Feel free to add to it and edit it.

PR #12, opened 2016-03-17
Add an initial section about limits

certik:
I found a good intro to limits that I wrote long time ago here: https://github.com/sympy/svn_materials/blob/4c79cfc68a85a947c5d730ec57540b6dff322f2d/presentations/SD6.pdf

### DIFF
--- a/algorithms.tex
+++ b/algorithms.tex
@@ -27,6 +27,63 @@ In [2]: limit((2*E**((1-cos(x))/sin(x))-1)**(sinh(x)/atan(x)**2), x, 0)
 Out[2]: E
 \end{verbatim}
 
+\subsubsection{Details}
+
+We first define comparability classes by calculating $L$:
+\begin{equation}
+L\equiv \lim_{x\to\infty} {\log |f(x)| \over \log |g(x)|}
+\end{equation}
+And then we define the $<$, $>$ and $\sim$ operations as follows: $f>g$ when
+$L=\pm\infty$ ($f$ is more rapidly varying than $g$, i.e. $f$ goes to $\infty$
+or $0$ faster than $g$, $f$ is greater than any power of $g$), $f<g$ when $L=0$
+($f$ is less rapidly varying than $g$) and $f\sim g$ when $L\neq 0,\pm\infty$
+(both $f$ and $g$ are bounded from above and below by suitable integral powers
+of the other).
+
+Examples:
+$$2 < x < e^x < e^{x^2} < e^{e^x}$$
+$$2\sim 3\sim -5$$
+$$x\sim x^2\sim x^3\sim {1\over x}\sim x^m\sim -x$$
+$$e^x\sim e^{-x}\sim e^{2x}\sim e^{x+e^{-x}}$$
+$$f(x)\sim{1\over f(x)}$$
+
+The Gruntz algorithm, on an example:
+$$f(x) = e^{x+2e^{-x}} - e^x + {1\over x}$$
+$$\lim_{x\to\infty} f(x) = ?$$
+
+Strategy:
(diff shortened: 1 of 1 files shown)


certik:
After #11 is resolved, I'll fix the citation.

certik:
@moorepants what's the best setup, so that the PR gets built and people can check the pdf that it looks good before merging?

asmeurer:
Sharelatex only seems to build the pdf on master. We could build the latex on Travis (that would at least verify that it builds without errors). It would be better to use some other CI, though, as I think this will share the Travis queue with SymPy.

certik:
Yes, it will. We should have left the repository under your name.

asmeurer:
Maybe if we use branches on this repo, then sharelatex will build them.

paper.tex
```diff
@@ -4,5 +4,18 @@
 \begin{document}
 \section{Introduction}
```
asmeurer, line 6: Need `\section{Features}` here.

asmeurer:
To make it easier to collaborate on this, would it be better to split the different sections into different files? I've never real-time collaborated on a LaTeX document on GitHub before, so I don't know what the best way to do things is.

moorepants:
The initial commits are a pain. It's usually good to have someone to start the document with sections or some base content. Then it also helps to assign people to sections for the first drafts. Then the editing with github is easier after that.

Another option is to enforce a rule where you have to put a carriage return in the tex source after each sentence. Then diffs play more nicely.

asmeurer:
Look good? https://github.com/sympy/sympy-paper/pull/15/files

certik:
I've rebased on top of the latest master.

leosartaj:
We also have an algorithm based on the paper "Computing limits of Sequences" by Manuel Kauers. Maybe we can also add an eg. for this. 

```
>>> limit(Sum(1/x, (x, 1, y)) - log(y), y, oo)
EulerGamma
```

certik:
@asmeurer would you mind reviewing this (or assigning somebody)? I put all my notes from [this](https://github.com/sympy/svn_materials/blob/4c79cfc68a85a947c5d730ec57540b6dff322f2d/presentations/SD6.pdf) old presentation. All the information is there, it just needs to be polished, so that it reads nicely. This can probably be done after it's merged into the first draft.

algorithms.tex
```diff
@@ -15,6 +15,76 @@ \subsection{The Risch Algorithm}
 
 \subsection{The Gruntz Algorithm}
 
+The limit modules implements the Gruntz algorithm
```
thilinarmtb: `limit module` or `implement`

algorithms.tex
```diff
@@ -15,6 +15,76 @@ \subsection{The Risch Algorithm}
...
+Out[2]: E
+\end{verbatim}
+
+\subsubsection{Details}
+
+We first define comparability classes by calculating $L$:
+\begin{equation}
+L\equiv \lim_{x\to\infty} {\log |f(x)| \over \log |g(x)|}
+\end{equation}
+And then we define the $<$, $>$ and $\sim$ operations as follows: $f>g$ when
```
asmeurer, line 36: Should this be `\lesssim` and `\gtrsim`?

certik, line 36: I don't think so: `\lesssim` is ≲,  `\gtrsim` is ≳. I want `<` and `>`.

algorithms.tex
certik: Thanks, fixed in c1e7e0ec57506ce2768ecb5274657a10cbe61ced.

moorepants:
I'm going to go ahead and merge this as it seems to be ready.

certik:
Thanks @moorepants.

PR #13, opened 2016-03-17
Testing (ignore)

asmeurer:
Testing if sharelatex will build this.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -4,5 +4,6 @@
 \begin{document}
 \section{Introduction}
 
+Testing
 
 \end{document}


asmeurer:
It's there https://www.sharelatex.com/github/repos/sympy/sympy-paper/. It would be nice if it registered builds with GitHub so that they showed up on the pull request.

asmeurer:
Oh, maybe it's not actually there. I can't get sharelatex to work. I think maybe the repository move confused it. If I try to turn on the hook I get an error message.

PR #27, opened 2016-03-31
Add basic outline for the rest of the paper

asmeurer:
Before we go any further, we need to agree on this outline. Once we have, I can assign writing to different people (I have made a list of who has volunteered to write what on [the wiki](https://github.com/sympy/sympy-paper/wiki/Who-Will-Write-What)). I'd like to get writing assignments out by Monday, so please comment on this before then.

### DIFF
--- a/algorithms.tex
+++ b/algorithms.tex
@@ -4,11 +4,14 @@
 % The sections here are preliminary. We may end up needing to cut some of
 % this.
 
+% XXX: Perhaps this should just be integrated into the features section.
+
 \subsection{Numerics}
 
 \subsection{Polynomials}
 
 \subsection{The Risch Algorithm}
+% Also the Meijer-G algorithm, if someone can write about it
 
 \subsection{The Gruntz Algorithm}
 
--- a/algorithms.tex
+++ b/algorithms.tex
@@ -1,2 +1,17 @@
+% A description of some of the algorithms in SymPy. The list is not
+% exhaustive.
 
+% The sections here are preliminary. We may end up needing to cut some of
+% this.
 
+\subsection{Numerics}
+
+\subsection{Polynomials}
+
+\subsection{The Risch Algorithm}
+
+\subsection{The Gruntz Algorithm}
+
+\subsection{Logic}
+
+\subsection{Other}
--- a/architecture.tex
+++ b/architecture.tex
@@ -85,6 +85,8 @@ one can wrap one of the integers with \texttt{Integer}, like \texttt{x +
 downside to using Python directly instead of a custom DSL, and we consider it
 to be worth it for the advantages listed above.
 
+\subsection{The Core}
+
 %%
 %% Assumptions
 \subsection{Assumptions}
--- a/architecture.tex
+++ b/architecture.tex
@@ -87,10 +87,14 @@ to be worth it for the advantages listed above.
 
 %%
 %% Assumptions
+\subsection{Assumptions}
+
 %%
 %% Extensibility
+\subsection{Extensibility}
 
 %% Functions
+\subsection{Functions}
 
 Mathematical functions are implemented as Python classes.
 
@@ -102,5 +106,3 @@ Mathematical functions are implemented as Python classes.
 %%%% Rewriting rules.
 
 %% Withheld evaluated (appears frequently on StackOverflow as a question).
-
-
--- a/features.tex
+++ b/features.tex
@@ -1,20 +1,21 @@
 % Features to discuss in-depth:
 
 % Basic operations (the core)
+\subsection{Basic Operations}
+% Including the assumptions system
 
-% Calculus (differentiation, integration, limits). Integration and limits in particular have nontrivial algorithms behind them.
+% Calculus (differentiation, integration, limits). Note that algorithm
+% descriptions will go in the algorithms section.
+\subsection{Calculus}
 
 % Solvers (regular equations, maybe also mention other types of solvers like ODEs/recurrence/Diophantine)
+\subsection{Solvers}
 
 % Matrices (worth including to stress that they are symbolic)
-
-% Assumptions system (could also be part of the core)
-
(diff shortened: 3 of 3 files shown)


moorepants:
Looks like a good start to me! Thanks for putting that together. It will help us get stared easier.

architecture.tex
```diff
@@ -87,10 +87,14 @@
...
 %%
 %% Assumptions
+\subsection{Assumptions}
+
 %%
 %% Extensibility
+\subsection{Extensibility}
 
 %% Functions
+\subsection{Functions}
```
leosartaj, line 99: I can help out with how functions are implemented/represented in SymPy.

algorithms.tex
```diff
@@ -1,2 +1,17 @@
+% A description of some of the algorithms in SymPy. The list is not
```
moorepants, line 1: Should Algs be with the description of the respective feature instead of in a stand alone section?

algorithms.tex
```diff
@@ -1,2 +1,17 @@
...
+
+\subsection{Polynomials}
+
+\subsection{The Risch Algorithm}
+
+\subsection{The Gruntz Algorithm}
+
+\subsection{Logic}
+
+\subsection{Other}
```
scopatz, line 20: Other probably isn't specific enough.

architecture.tex
```diff
@@ -87,10 +87,14 @@
```
moorepants, line 87: How about a section to describe the args invariant, the tree structure, and the fundamental classes.?

features.tex
```diff
@@ -22,3 +23,4 @@
 % combinatorics/group theory, code generation, lie algebras, cryptography,
 % category theory, special functions, sets, matrix expressions, series, or
 % vectors.
+\subsection{Other features}
```
leosartaj, line 26: Series, in particular about Formal Power Series. Also, can help out with the current series expansion infrastructure.(I do not have a lot of idea about `rs_series` which should/can be talked about.)

scopatz:
My only comment is that the other sections should likely be removed or mention specifically what they aim to discuss.

architecture.tex
asmeurer, line 87: Good idea.

asmeurer, line 87: Although I don't know what to call it.

algorithms.tex
asmeurer, line 1: Maybe. We would have to rethink how the feature list looks like. Right now it's just a section for each of the most core features, and then a short listing of the rest of the features. 

I'm actually most unsure about this algorithms section to be honest.

asmeurer:
The "other features" section will literally be a listing of every feature listed in the comment above the section. 

For algorithms, I guess I was thinking the same thing. But I'm really unsure what the whole section should look like. 

At any rate, you're probably right that "other" is not the best word.

scopatz:
Well, just a listing of features probably does not deserve it own section. That could just be a table somewhere.

architecture.tex
moorepants, line 87: "Core Architecture"

"Core Class Hierarchy"

ashutoshsaboo:
Can we view the commits, in some PDF format / in some rendered LaTeX format? @asmeurer 

Also, I am not sure but, do we also add the SymPy Live and SymPy Gamma section in this outline, since they'll also be a part of this paper right?

asmeurer:
Sharelatex doesn't seem to work so well, and I never got Travis so that it uploads builds somewhere, so for now, you'll have to build it yourself to see the pdf.

algorithms.tex
```diff
@@ -1,2 +1,17 @@
...
+% exhaustive.
 
+% The sections here are preliminary. We may end up needing to cut some of
+% this.
 
+\subsection{Numerics}
+
+\subsection{Polynomials}
+
+\subsection{The Risch Algorithm}
```
certik, line 13: Also MeijerG algorithm.

certik:
It looks good.

(Btw, as to uploading build artifacts, that's where GitLab is already taking over both GitHub and Travis: http://doc.gitlab.com/ce/ci/build_artifacts/README.html, i.e. their CI can upload the artifacts and get it served by GitLab itself. So a PR would upload the pdf and anybody can download it easily for each commit.)

algorithms.tex
asmeurer, line 13: I think it would be good to have, but only if someone can write about it. Otherwise, we should just mention that it's there without describing it, along with the other algorithms that we just mention.

algorithms.tex
shivamvats, line 1: I too think it makes more sense to mention the algorithms in the respective feature. Since most of our algorithms are from literature, we can describe modifications/optimizations that we did (if any).

algorithms.tex
ashutoshsaboo, line 20: The outline seems pretty good me. But just one thing, instead of mentioning the 'Other' heading as discussed earlier, it would rather be good if we first list all the algorithms, before all the headings,i.e, before **0.1** , and then we can continue describing the topics that have been mentioned in the points **0.1 - 0.5** as of now. It would rather make the structure of the outline better I guess. How about this? Any opinions? :smiley: @asmeurer

ashutoshsaboo:
Another thing, since we are working on our outline. So, how about if we also think about the major points that need to be listed in each section, say that of Algorithms as well? It will make our work much more organized and easier as well. 

As, in like, in "Algorithms" section, we can have atleast these points definitely-:
1. Description about the algorithm in general
2. How does SymPy manage to implement it with some illustrative examples. 

How about adding these details as well if possible? Any opinion on this?

algorithms.tex
asmeurer, line 20: That's a good point. I'm not worried about organizational issues right now. My main concern is getting the stuff written. Then we can arrange it. But you're probably right that it would be better to start with the generic list and then have specific sections on things we want to expand on.

PR #34, opened 2016-04-11
numerics and mpmath

fredrik-johansson:
I added some text about numerics and mpmath. I hope the mpmath part is not too much. Could add a few more concrete examples. Are there any basic concepts that need elaboration?


algorithms.tex
```diff
@@ -8,6 +8,143 @@
...
+and use algorithms that support computig hundreds of digits rapidly,
+except in degenerate cases.
+
+The double exponential (tanh-sinh) quadrature is used for numerical
+integration by default. For smooth integrands, this algorithm usually
+converges extremely rapidly, even when the integration interval is infinite
+or singularities are present at the endpoints.
+However, for good performance, singularities
+in the middle of the interval must be specified
+the user.
```
asmeurer: _by_

algorithms.tex
```diff
@@ -8,6 +8,143 @@
...
+and the redundant integer $b$ stores the bit length of $x$ for quick access.
+If GMPY [citation?] is installed, mpmath automatically switches to
+using the \texttt{gmpy.mpz} type for $x$ and using GMPY helper methods
+to perform rounding-related operations, improving performance.
+
+The mpmath library includes support for complex numbers, evaluation of
+special functions, root-finding, linear algebra, polynomial approximation,
+and numerical computation of limits, derivatives, integrals, infinite
+series, and ODE solutions. All features work in arbitrary precision
+and use algorithms that support computig hundreds of digits rapidly,
```
asmeurer: computing

algorithms.tex
```diff
@@ -8,6 +8,143 @@
...
+hypergeometric-type linear combinations of the above type.
+The implementation automatically detects cancellation problems,
+and computes limits numerically by perturbing parameters whenever
+internal singularities occur (the perturbation size is automatically
+decreased until the result is detected to converge numerically).
+
+It is important for a computer algebra system
+to support generalized hypergeometric functions and
+Meijer G-functions robustly, due to their frequent appearance
+in closed forms for integrals and sums.
```
asmeurer: Would be worth pointing out at this point that SymPy does in fact implement algorithms for computing integrals in terms of Meijer G-functions.

algorithms.tex
```diff
@@ -8,6 +8,143 @@
...
+uncertainty in measurement data in any meaningful way;
+the sole purpose is to mitigate loss of accuracy that typically occurs
+when converting symbolic expressions to numerical values, for example
+due to catastrophic cancellation. This is illustrated by the following
+example:
+
+\begin{verbatim}
+>>> cos(exp(-100)).evalf(25) - 1
+0
+>>> (cos(exp(-100)) - 1).evalf(25)
```
asmeurer, line 44: You don't state anywhere that `evalf(25)` means to use 25 decimal digits of precision (although it may be obvious).

asmeurer:
This looks good. Is there an mpmath citation that we should include?

algorithms.tex
asmeurer: Or if someone is able to write about that in more detail in the algorithms/features section we can just cross reference that.

fredrik-johansson:
Indeed, my intention was to crossref the symbolic integration section, when/if there is one. I definitely think there should be one, but I don't think I should be writing it...

I added a bit more: lambdify/codegen, remark on Mathematica, simple mpmath usage example, more on the Meijer G-function, and references.

We could add a very basic code example at the start, showing how to add two Float(...) values, say.

It's over three pages now; I hope it's not too much already. I could obviously add a lot more, but better not. If you want to take something out, I won't mind.

fredrik-johansson:
There's no mpmath ref. We could site the software itself with the url to the website, of course.

GMPY and NumPy are also mentioned now.

asmeurer:
My guess is that we'll have to end up cutting from all the sections, but I want to see what all we have from everybody before we do that.

algorithms.tex
```diff
@@ -34,18 +34,70 @@ \subsection{Numerics}
...
-[todo: contrast with mathematica's significance arithmetic? cite Sofroniou2005precise, maybe Fateman]
+\begin{verbatim}
+>>> f = lambdify((x, y), sin(x*y)**2, modules='numpy')
+>>> from numpy import array
+>>> f(array([1,2,3]), array([4,5,6]))
+array([ 0.57275002,  0.29595897,  0.56398184])
+\end{verbatim}
+
+SymPy can also generate C, C++, Fortran77, Fortran90 and
+Octave/Matlab source code, via the \texttt{codegen} function. [document this?]
```
asmeurer, line 86: I would rather mention the functions `ccode`, `fcode`, etc.

algorithms.tex
```diff
@@ -34,18 +34,70 @@ \subsection{Numerics}
...
+numbers in the rest of the system.
+
+\subsubsection{Code generation}
+
+SymPy's \texttt{lambdify} can be used to convert a symbolic expression to a
+callable Python function for faster numerical evaluation.
+Various back ends are supported. The following example
+demonstrates creating a NumPy-based function from
+a SymPy expression, which automatically supports
+vectorized array evaluation [NumPy citation?]:
```
asmeurer: https://www.scipy.org/citing.html#numpy-scipy

algorithms.tex
asmeurer, line 86: Code generation is mentioned in the table at https://github.com/sympy/sympy-paper/pull/42, which should be sufficient. 

The primary example in this section is `lambdify`, which is technically code generation, but probably not what first comes to mind for most people. A better goal for this section might be interaction with other Python libraries (particularly numpy).

algorithms.tex
```diff
@@ -8,6 +8,220 @@
...
+The \texttt{evalf} method works with complex numbers and supports
+more complicated expressions, such as
+special functions, infinite series and integrals.
+
+SymPy does not track the accuracy of
+approximate numbers outside of \texttt{evalf}.
+The familiar dangers of floating-point arithmetic apply, and 
+symbolic expressions containing floating-point numbers should be treated
+with some caution.
+This approach is similar to Maple and Maxima.
```
asmeurer, line 57: Citations?

algorithms.tex
```diff
@@ -8,6 +8,220 @@
...
+-6.919482633683687653243407e-88
+\end{verbatim}
+
+The \texttt{evalf} method works with complex numbers and supports
+more complicated expressions, such as
+special functions, infinite series and integrals.
+
+SymPy does not track the accuracy of
+approximate numbers outside of \texttt{evalf}.
+The familiar dangers of floating-point arithmetic apply, and
```
asmeurer: Is there a good citation for the "familiar dangers of floating point arithmetic"?

moorepants: Yes, I've read a really nice expose once. Let me see if I can find it.

moorepants: It may be this: http://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

asmeurer:
It may be worth mentioning somewhere that floating point literals are evaluated by Python into machine floats, so it's preferable to create `Float` or `mpf` objects with strings to avoid unnecessary precision loss.

asmeurer:
I think this looks mostly good now, aside from those comments. If you can't find good citations now just add some comments so we find them later.

moorepants:
@fredrik-johansson Can you address Aaron's comments so we can get this merged?

certik:
Also you need to merge with the latest master to resolve conflicts.

algorithms.tex
fredrik-johansson, line 57: What do you mean?

asmeurer, line 57: We will at least need general citations for Maple and Maxima. If there are more general ones for the fact that they use this approach to floating point numbers that would be nice as well.

PR #35, opened 2016-04-11
Basic operations

Upabjojr:


asmeurer:
There's some overlap with what you have on the assumptions and what I have at https://github.com/sympy/sympy-paper/pull/36. We'll need to deduplicate things, but make sure things are introduced properly before they are discussed.

Upabjojr:
Damn, that's a total overlap. I saw the assignment about the basic operations, and wrote this up. How do you plan to deal with this?

features_basic_operations.tex
```diff
@@ -0,0 +1,153 @@
+
+\subsubsection{Assumptions system}
+
+% Including the assumptions system
+
+
+SymPy has two assumptions systems, referred to as new-style and old-style assumptions.
+
+In the old-style assumptions system prepositions are assigned to symbols
```
asmeurer: propositions

features_basic_operations.tex
```diff
@@ -0,0 +1,153 @@
...
+upon class construction, for example, to declare the symbol $i$ as positive integer,
+one would call
+
+\begin{verbatim}
+i = Symbol("i", integer=True, positive=True)
+\end{verbatim}
+
+querying the assumptions is handled through class methods
```
asmeurer: I would call these "attributes". "Class method" has a specific meaning in Python that doesn't apply here.

features_basic_operations.tex
```diff
@@ -0,0 +1,153 @@
...
+negative.
+
+The new-style assumptions are an assumptions system that exists alongside with
+the old-style, but is significantly different in the way predicates are
+used.
+Predicates in the new-style assumptions system are located under the \textit{Q}
+namespace, they appear as
+\verb|Q.positive|, \verb|Q.integer| and so on,
+and, as of the writing of this paper, 
+they can accept only one parameter.
```
asmeurer: For the existing predicates, only one parameter makes sense.

asmeurer:
I'm unsure what to do about the duplicate work on assumptions. Maybe others have thoughts. What I wrote is directed as an internals discussion, and what you wrote is directed as usage. Some possible suggestions
- Move the features section before the architecture section, and then rewrite the architecture section under the assumption that the features themselves were already introduced. 
- Leave what I have in the architecture section and remove the assumptions from this section.

Upabjojr:
Maybe my introduction to the assumptions is a bit more _introductive_, without introducing the internals immediately.

Shall I concentrate on the API usage in the description?

asmeurer:
My thoughts were that the features section would focus on usage, and the internals would be discussed in other sections (architecture or algorithms). Others have suggested merging the two. I'm unclear which way is better, but I am starting to think that if we separate them, the features section should go before the other two sections.

features.tex
```diff
@@ -2,10 +2,9 @@
 
 % Basic operations (the core)
 \subsection{Basic Operations}
-% Including the assumptions system
```
asmeurer, line 5: If we move the features section to the beginning of the paper, we should include some description of how to define symbols and create expressions.

moorepants:
This looks fine to merge as a first draft.

moorepants:
I'll merge later today unless there are objections.

PR #37, opened 2016-04-12
Add a section about Solvers

aktech:
Solvers section added.

### DIFF
--- a/solvers.tex
+++ b/solvers.tex
@@ -1,15 +1,20 @@
 %% Solvers in SymPy
 
-There are two submodules to solve equations in SymPy, referred to as old solve
-i.e. \texttt{solve} and new solve i.e. \texttt{solveset}. There are several
-design changes in \texttt{solveset} to resolve the issues with old solve, for
-example old \texttt{solve}'s input API has many flags which are not needed and
-they make it hard for the user and the developers to work on solvers. In
-contrast to old solve, the \texttt{solveset} has a clean input API, It only
-asks for the much needed information from the user, following are the function
-signatures of old and new solve:
+SymPy has a solvers module to solve equations symbolically. There are two
+submodules to solve equations in SymPy, referred to as old solve i.e.
+\texttt{solve} and new solve i.e. \texttt{solveset}. Solveset is introduced
+with several design changes with respect to old \texttt{solve} to resolve the
+issues with old \texttt{solve}, for example old \texttt{solve}'s input API has
+many flags which are not needed and they make it hard for the user and the
+developers to work on solvers. In contrast to old solve, the \texttt{solveset}
+has a clean input API, It only asks for the much needed information from the
+user, following are the function signatures of old and new solve:
 
 \begin{verbatim}
     solve(f, *symbols, **flags)  # Old Solve
     solveset(f, symbol, domain)  # New Solve
 \end{verbatim}
+
+The old \texttt{solve} has an inconsistent output API for various types of
+inputs, whereas the \texttt{solveset} has a canonical output API using sets.
+It can consistently output various types of solutions.
--- a/solvers.tex
+++ b/solvers.tex
@@ -1,20 +1,66 @@
 %% Solvers in SymPy
 
+
 SymPy has a solvers module to solve equations symbolically. There are two
-submodules to solve equations in SymPy, referred to as old solve i.e.
-\texttt{solve} and new solve i.e. \texttt{solveset}. Solveset is introduced
-with several design changes with respect to old \texttt{solve} to resolve the
-issues with old \texttt{solve}, for example old \texttt{solve}'s input API has
-many flags which are not needed and they make it hard for the user and the
-developers to work on solvers. In contrast to old solve, the \texttt{solveset}
-has a clean input API, It only asks for the much needed information from the
-user, following are the function signatures of old and new solve:
+submodules to solve algebraic equations in SymPy, referred to as old solve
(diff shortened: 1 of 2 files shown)


solvers.tex
```diff
@@ -0,0 +1,66 @@
...
+\begin{verbatim}
+>>> solveset(x - 1)
+>>> {1}
+\end{verbatim}
+
+\begin{itemize}
+\item Finite set of solution, quadratic equation
+\end{itemize}
+\begin{verbatim}
+>>> solveset(x**2 - pi**2, x)
```
aktech, line 36: I am yet to figure out, how to left align code inside itemize, so as to wrap all of them in one itemize.

isuruf:
Should we mention both `solve` and `solveset`? Since both methods are introduced in this paper, I think it should only talk about `solveset` to avoid confusing the readers

aktech:
@isuruf I was also skeptical about adding both, but the problem is the `solveset` is not as good as `solve` as of now. So it may be worth mentioning some features of solve, which solveset is incapable of. isn't it?

solvers.tex
```diff
@@ -0,0 +1,66 @@
...
+\begin{itemize}
+\item Interval of solution
+\end{itemize}
+\begin{verbatim}
+>>> solveset(x**2 - 3 > 0, x, domain=S.Reals)
+(-oo, -sqrt(3)) U (sqrt(3), oo)
+\end{verbatim}
+
+\begin{itemize}
+\item Infinitely many solution
```
ashutoshsaboo: Infinitely many solutions

ashutoshsaboo:
I agree in a way with @aktech , @isuruf . But, what I suggest, is that, more examples need to be added also corresponding to `solve` - those examples which show the functionality that currently `solveset` is not able to achieve and `solve` achieves easily. Also, I feel the `solve` and `solveset` doesn't create too much confusion as well, since @aktech managed to clearly mention the differences between those two in his draft. Your opinion @isuruf ?

I guess corresponding to your `solveset` examples, if we can add examples that show the inconsistency output API of `solve`, then it'd be good I guess. That is, I guess we must give examples to support the functionality of both `solve` and `solveset`. 

Your opinion on this? @aktech 

cc/ @asmeurer

aktech:
@ashutoshsaboo +1

ashutoshsaboo:
Current draft seems to be okay with me. I would suggest you @aktech , to add those relevant examples (that I referred to in my previous comment), then we may proceed ahead. 

:smile: cc/ @asmeurer

aktech:
>  if we can add examples that show the inconsistency output API of solve

I am not sure, if that would be helpful.

moorepants:
Wouldn't it be better to only show solveset in this paper, being that the old solve will eventually be removed? I don't see how explaining differences in the two is that important for the paper. All we really need the reader to know is that SymPy can solve equations and what types of equations SymPy can solve. I don't think it will benefit the reader of the paper to know the complex details about the solve and solveset relationship.

aktech:
> Wouldn't it be better to only show solveset in this paper, being that the old solve will eventually be removed? 

I think we should let people know that SymPy is capable of solving non-linear system and transcendental equations too and to show that we need to use `solve` examples and It may confuse people, that why have two functions to solve equations? that's why I have written some background about the evolution of solveset from solve, so that people know that currently what they can do with `solve` and `solveset`combined , they can do the same with `solveset` alone in future after the replacement.
ss

moorepants:
I like that table a lot. That presents things clearly.

Just out of curiosity (not for this paper) but could simply have solveset call solve and wrap the result in a set for the last two?

aktech:
> Just out of curiosity (not for this paper) but could simply have solveset call solve and wrap the result in a set for the last two?

That's a good question, but that's a temporary solution since `solveset` is designed to consider some other important factors like guarantee of returning correct solution, all the solutions in the required domain. The current heuristics works for a lot of cases, but we need to re-implement it such that to follow these basic principles, and of-course we can take a lot from the old solve ;)

solvers.tex
```diff
@@ -0,0 +1,127 @@
+%% Solvers in SymPy
+
+
+SymPy has a solvers module to solve equations symbolically. There are two
```
Upabjojr: _module of equation solvers for symbolic equations_ sounds better.
I would avoid repeating _solvers_, _solve_ two times.

solvers.tex
```diff
@@ -0,0 +1,127 @@
...
+submodules to solve algebraic equations in SymPy, referred to as old solve
```
Upabjojr, line 5: _old_ or _old-style_ ?

Upabjojr, line 5: _old-style solve**r**_  maybe?

solvers.tex
aktech: +1

solvers.tex
aktech, line 5: In SymPy the `solve` function is referred to as old solve, in the documentation of [solveset](http://docs.sympy.org/latest/modules/solvers/solveset.html) (the new solve), so, I think its a good idea to use same convention, isn't it?

Upabjojr, line 5: OK. Grammar problem remains, _solve_ is a verb, verbs cannot have adjectives or used as complements (unless -ing form is added). Maybe _old solve function_?

aktech, line 5: > Maybe _old solve function?_

+1

moorepants:
Please address the final comments so that we can merge today.

aktech:
@moorepants done.

PR #44, opened 2016-04-13
Introduction

asmeurer:
Does it seem like a good intro? What else should be mentioned here?


moorepants:
This makes me want to step back and talk about what the "theme" of the paper is. Being that we are submitting this as an academic publication, it would be worth thinking about what the academic contribution of SymPy is. If the paper is strictly a description of a particular software, I'm not sure typical journals will be that fond of accepting it (have we chosen a journal?). The intro can set the tone for the theme. Another good question I like to ask when writing papers is "What story do we want to tell?".

It isn't yet clear to me what story we are telling. It would be nice if we can show how SymPy solves particular problems better than other CASs, or is more suitable for scientific work, or that community driven CAS development creates software that better suites scientific computing needs.

Have we had any conversation about this yet? I may have missed it.

This intro could help set the tone of all the other sections.

introduction.tex
```diff
@@ -1,2 +1,40 @@
+SymPy is an full featured computer algebra system (CAS) written in the Python
```
Upabjojr: an ===> a

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+programming language. It is open source, licensed under the extremely
+permissive BSD license, which allows anyone to reuse SymPy or its source code,
```
Upabjojr: which of the various BSD licenses?

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+even for commercial purposes. SymPy was started by Ond\v{r}ej \v{C}ert\'{\i}k
+in 2005, and it has since grown into a large open source project, with over
+500 contributors.
+% citation?
 
+SymPy is written entirely in the Python programming language,
```
Upabjojr: mention SymEngine?

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+% cite Python?
+which is also the language used to interact with it, both programmatically and
+interactively. Python is a popular dynamically typed programming language that
+has a focus on ease of use and readability. Unlike many CASs, SymPy does not
+invent its own programming language. We rather take the approach that Python
```
Upabjojr: We ===> It

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+is already a well-designed programming language, built by expert language
+designers, something which the majority of SymPy developers are not. Python is
```
Upabjojr: Maybe avoid subjective judgments.

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+also a very popular language for scientific computing and data science, with a
+wide range of useful libraries.
```
Upabjojr: Mention the potentiality of community growth by using Python, given that Python is on the top 10 programming languages in most rankings.

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+% Cite numpy, scipy, pandas
+% We could also cite
+% https://stackoverflow.com/research/developer-survey-2016#most-popular-technologies-per-occupation
 
+SymPy is designed with a strong focus that it be usable as a library. This
+means that extensibility is important in its API design. This is also one of
+the reasons SymPy makes no attempt to extend the Python language itself. The
```
Upabjojr: we have `isympy` and `ipython`...

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+goal is for users of SymPy to be able to import SymPy alongside other Python
+libraries in their workflow, whether that is an interactive workflow or
+programmatic use as part of a larger library. This does have some downsides;
+for instance, unlike many popular CASs, all variables in SymPy must be defined
```
Upabjojr: acronym plural, _CAS's_ or _CASs_?

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+before they are used (Python does not provide a mechanism to automatically
+declare variables). This is a minor inconvenience for interactive use, but we
+consider it to be good practice for any reproducible work, and it aligns with
```
Upabjojr: Misspellings of variable names in Mathematica are a common cause of headache, as Mathematica accepts them as valid, I wouldn't consider this an inconvenience.

moorepants: I also don't think this belongs in the introduction. It seems like more of a technical detail.

introduction.tex
asmeurer: From what I can tell, the consensus is that an apostrophe is wrong, because it's not possessive. See https://english.stackexchange.com/questions/503/what-is-the-correct-way-to-pluralize-an-acronym.

asmeurer:
Yeah, OK, let's discuss what the introduction should be.  I just sat down and wrote what I felt should go here, but I think it could be better. 

The point of the paper is to be _the_ paper for SymPy, i.e., this will be the paper that people cite when they use SymPy. 

To that end, I feel it should primarily answer two questions: "what is SymPy?" and "why were the various aspects of its design done the way they were?"

So my goal in writing this was to give high level answers to these questions. The rest of the paper will go into more detail. For example: what is SymPy? It's an open source computer algebra system written in Python. Why is it written in Python? I tried to answer that here. What are the core principles of its design, and why? I also attempted to hit this here, with at least one core principle (usability as a library). I may have gone off too much on a tangent with some nuances from choosing Python (it probably should be moved to the architecture section). 

What are the other _core_ design principles, which deserve discussion in the introduction? I'm not sure. "Using Python" and "being a library" are two that come to mind, but I may have missed some. Perhaps the goal of being "full featured" deserves some further discussion. Also, [this](https://github.com/sympy/sympy/wiki/The-Zen-of-SymPy) can be inspiration. 

I do think that we collectively are not really agreed on the theme of the paper, which is clear from the conflicting desires of the various pull requests. So what do others think the theme should be? Do you agree with my assessment above?

moorepants:
Aaron, I like what you say above:
- Should be _the_ paper for SymPy.
- Should answer "What is SymPy?"
- Should answer "Why were the various aspects of design done the way they were?"

I'd add these:
- Should answer "Why is SymPy useful to researchers, educators, and companies?"
- Should answer "How does SymPy compare to other similar software and why should you use it instead of or alongside them?"

I think these two are a bit more important wrt to _the_ sympy paper than the design decisions. The design decisions are more interesting to designers of CAS systems and less so for users.

One question to ask ourselves:

Who is the intended audience?

Finally, I was thinking that we could write the abstract first. If we can agree on an abstract it will frame the "mission" of the paper. Then when each person writes their section, they can refer to the mission to judge whether what they are writing fits the theme.

aktech:
> - Should answer "Why is SymPy useful to researchers, educators, and companies?"
> - Should answer "How does SymPy compare to other similar software and why should you use it instead of or alongside them?"
> 
> I think these two are a bit more important wrt to the sympy paper than the design decisions. The design decisions are more interesting to designers of CAS systems and less so for users.

@moorepants I cannot agree with you more, here. +1

introduction.tex
ashutoshsaboo: Yes, CASs is the correct plural here. :+1:

ashutoshsaboo:
Hi,

I somewhat agree with @moorepants . It'll be very useful, if we mention the way SymPy helps educators, researchers, and companies, as Mr. Moore suggested. It would also provide the targeted audience a reason, as to why to use SymPy if we write about the above. So, I guess, the addition of the 2 points that Mr. Moore suggested above, would make our introduction, a pretty good one @asmeurer . 

In that way, we could also easily build up for explaining all the full, extensive features list of SymPy as well.

But, I guess we must also **not miss upon** the important architecture details of SymPy as a CAS, and I feel that, they must only be mentioned in the `Architecture` section, and not the `Introduction`. @asmeurer

moorepants:
@asmeurer Can you fix @Upabjojr's comments?

We can move the theme discussion to a new issue.

introduction.tex
scopatz: I don't think that is necessary info in the intro right up front.

introduction.tex
```diff
@@ -1,2 +1,40 @@
+SymPy is a full featured computer algebra system (CAS) written in the Python
+programming language. It is open source, licensed under the extremely
+permissive 3-clause BSD license, which allows anyone to reuse SymPy or its source code,
+even for commercial purposes. SymPy was started by Ond\v{r}ej \v{C}ert\'{\i}k
+in 2005, and it has since grown into a large open source project, with over
+500 contributors.
```
scopatz: This sentence sounds more like an advertisement.  I would get rid of it or move it later in the paper. This isn't supposed to be a paper on the community, per se.  And @certik is an author of the paper so it is weird to mention his name here.

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+% citation?
 
+SymPy is written entirely in the Python programming language,
+% cite Python?
```
scopatz, line 14: Yep, citation needed.

introduction.tex
scopatz: Yeah, I'd kill the last couple of "clauses".  Rather I'd emphasize that this is about code reuse and reduction of effort and eases the learning curve for new developers. The value one places on language design is subjective.

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+interactively. Python is a popular dynamically typed programming language that
+has a focus on ease of use and readability. Unlike many CASs, SymPy does not
+invent its own programming language. We rather take the approach that Python
+is already a well-designed programming language, built by expert language
+designers, something which the majority of SymPy developers are not. Python is
+also a very popular language for scientific computing and data science, with a
+wide range of useful libraries.
+% Cite numpy, scipy, pandas
+% We could also cite
+% https://stackoverflow.com/research/developer-survey-2016#most-popular-technologies-per-occupation
```
scopatz: Citations definitely needed.

introduction.tex
scopatz: I think the dynamic portions don't need to be discussed here.  Maybe mentioned as part of the tools sections.

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
 
+SymPy is designed with a strong focus that it be usable as a library. This
+means that extensibility is important in its API design. This is also one of
+the reasons SymPy makes no attempt to extend the Python language itself. The
+goal is for users of SymPy to be able to import SymPy alongside other Python
+libraries in their workflow, whether that is an interactive workflow or
+programmatic use as part of a larger library. This does have some downsides;
+for instance, unlike many popular CASs, all variables in SymPy must be defined
+before they are used (Python does not provide a mechanism to automatically
+declare variables). This is a minor inconvenience for interactive use, but we
```
scopatz: Again, I think we should refrain from making subjective statements.  This isn't a down side per se. It is a design decision. By analogy, declaring variables in most type safe languages is just part of the definition of these languages. Different people feel differently about it, but their feeling are largely irrelevant to a discussion of what the language definition is.

introduction.tex
```diff
@@ -1,2 +1,40 @@
...
+consider it to be good practice for any reproducible work, and it aligns with
+the philosophy of Python, ``explicit is better than implicit.''
+% cite https://www.python.org/dev/peps/pep-0020/
+
+SymPy does not have a built in graphical user interface (GUI), however, when
+used in the Jupyter Notebook
```
scopatz, line 39: Move the isympy discussion here.

introduction.tex
asmeurer: Maybe Ondrej's name shouldn't be mentioned, but I did want to highlight the community.

asmeurer:
It sounds like this will need to be rewritten.

introduction.tex
asmeurer: I agree. We don't mention SymEngine anywhere in the paper currently. It might be worth mentioning in the "conclusion and future work" section. SymEngine ought to have its own paper (if Ondrej wants to do that).

asmeurer:
OK, everyone, see my rewrite. I've tried to de-subjectify most of it. 

I also tried to expand a bit of the intro paragraph about the community, taking some bits from @moorepants's https://github.com/sympy/sympy-paper/pull/84, but it probably could use some work (feel free to suggest a rewrite).

asmeurer:
I have finished the introduction, and made other changes here as well. I will merge tomorrow if no one has any comments.

introduction.tex
certik: Yes, in fact we already started to work on a paper with @thilinarmtb and @isuruf. After the sympy paper is submitted, we will start a paper on symengine, and we will follow the exact same workflow (open on github, authorship criteria, etc.). Right now I want to concentrate on getting the sympy paper out.

PR #57, opened 2016-04-18
Added more details to the README.

moorepants:
I've added the CCBY license here and some other details. Let me know if that is cool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,32 @@
 # SymPy Paper
 
-[![Join the chat at https://gitter.im/asmeurer/sympy-paper](https://badges.gitter.im/asmeurer/sympy-paper.svg)](https://gitter.im/asmeurer/sympy-paper?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+This is the source repository for the paper:
+
+> SymPy Authors, "SymPy: Symbolic Computing in Python", 2016.
+
+This repository contains or links to all of the information needed to reproduce
+the results in the paper.
+
+The latest rendered version of the PDF can be viewed via the
+[ShareLaTeX](http://sharelatex.com) CI system:
 
 [![PDF Status](https://www.sharelatex.com/github/repos/asmeurer/sympy-paper/builds/latest/badge.svg)](https://www.sharelatex.com/github/repos/asmeurer/sympy-paper/builds/latest/output.pdf)
 
+# Contributing
 
-This a paper about SymPy.
+Join our chat room to discuss contributions:
+
+[![Join the chat at https://gitter.im/asmeurer/sympy-paper](https://badges.gitter.im/asmeurer/sympy-paper.svg)](https://gitter.im/asmeurer/sympy-paper?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
+# License
+
+The contents of this repository is licensed under the [Creative Commons
+Attribution 4.0 International
+License](http://creativecommons.org/licenses/by/4.0).
+
+[![CCBY](https://i.creativecommons.org/l/by/4.0/80x15.png)](http://creativecommons.org/licenses/by/4.0)
+
+# LaTeX Build Instructions
+
+Install a full LaTeX suite on your preferred operating system including minted
+and pygments. To build the PDF type `make` in the source directory.
--- a/README.md
+++ b/README.md
@@ -20,11 +20,15 @@ Join our chat room to discuss contributions:
 
 # License
 
-The contents of this repository is licensed under the [Creative Commons
-Attribution 4.0 International
-License](http://creativecommons.org/licenses/by/4.0).
-
 [![CCBY](https://i.creativecommons.org/l/by/4.0/80x15.png)](http://creativecommons.org/licenses/by/4.0)
+
+[![license](https://img.shields.io/github/license/mashape/apistatus.svg?maxAge=2592000)]()
+
+The contents of this repository are dual licensed under the [Creative Commons
+Attribution 4.0 International
+License](http://creativecommons.org/licenses/by/4.0) and the [MIT
+License](https://opensource.org/licenses/MIT). See the `LICENSE` file for
+details.
 
 # LaTeX Build Instructions
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,418 @@
+Attribution 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and does
+not provide legal services or legal advice. Distribution of Creative Commons
+public licenses does not create a lawyer-client or other relationship. Creative
+Commons makes its licenses and related information available on an "as-is"
+basis. Creative Commons gives no warranties regarding its licenses, any
+material licensed under their terms and conditions, or any related information.
+Creative Commons disclaims all liability for damages resulting from their use
+to the fullest extent possible.
+
+Using Creative Commons Public Licenses
+
+Creative Commons public licenses provide a standard set of terms and conditions
+that creators and other rights holders may use to share original works of
+authorship and other material subject to copyright and certain other rights
+specified in the public license below. The following considerations are for
+informational purposes only, are not exhaustive, and do not form part of our
+licenses.
+
+     Considerations for licensors: Our public licenses are
+     intended for use by those authorized to give the public
+     permission to use material in ways otherwise restricted by
+     copyright and certain other rights. Our licenses are
+     irrevocable. Licensors should read and understand the terms
+     and conditions of the license they choose before applying it.
+     Licensors should also secure all rights necessary before
+     applying our licenses so that the public can reuse the
+     material as expected. Licensors should clearly mark any
+     material not subject to the license. This includes other CC-
+     licensed material, or material used under an exception or
+     limitation to copyright. More considerations for licensors:
+     wiki.creativecommons.org/Considerations_for_licensors
+
+     Considerations for the public: By using one of our public
+     licenses, a licensor grants the public permission to use the
+     licensed material under specified terms and conditions. If
+     the licensor's permission is not necessary for any reason--for
+     example, because of any applicable exception or limitation to
+     copyright--then that use is not regulated by the license. Our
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,8 +1,32 @@
...
+
+[![Join the chat at https://gitter.im/asmeurer/sympy-paper](https://badges.gitter.im/asmeurer/sympy-paper.svg)](https://gitter.im/asmeurer/sympy-paper?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
+# License
+
+The contents of this repository is licensed under the [Creative Commons
+Attribution 4.0 International
+License](http://creativecommons.org/licenses/by/4.0).
+
+[![CCBY](https://i.creativecommons.org/l/by/4.0/80x15.png)](http://creativecommons.org/licenses/by/4.0)
```
asmeurer, line 24: We'll need to agree on this.

moorepants, line 24: How do you want to get agreement? Open a new new issue? Discuss here?

certik, line 24: I couldn't find if CC-BY 4.0 is an open source license. The only mention I found is here:

https://wiki.debian.org/DFSGLicenses#Creative_Commons_Attribution_Share-Alike_.28CC-BY-SA.29_v4.0

which says that a similar license CC-BY-SA 4.0 is compatible with DFSG, which is essentially the same as the open source definition at opensource.org. However, I found that a previous version of the license, CC-BY 1.0, is not open source:

https://wiki.debian.org/DFSGLicenses#Creative_Commons_Attribution_License_.28CC-by.29.2C_v1.0

So for these reasons, why cannot we just use an MIT or BSD license?

My vote is for MIT or BSD.

README.md
```diff
@@ -1,8 +1,32 @@
...
+The latest rendered version of the PDF can be viewed via the
+[ShareLaTeX](http://sharelatex.com) CI system:
 
 [![PDF Status](https://www.sharelatex.com/github/repos/asmeurer/sympy-paper/builds/latest/badge.svg)](https://www.sharelatex.com/github/repos/asmeurer/sympy-paper/builds/latest/output.pdf)
 
+# Contributing
+
+Join our chat room to discuss contributions:
+
+[![Join the chat at https://gitter.im/asmeurer/sympy-paper](https://badges.gitter.im/asmeurer/sympy-paper.svg)](https://gitter.im/asmeurer/sympy-paper?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
```
certik: I think we should rather use https://gitter.im/sympy/sympy-paper

moorepants: I think this was auto added by gitter. I can fix it.

README.md
moorepants, line 24: I don't know what you mean by "is this license open source"?

We can use one of the typical software licenses like MIT or BSD, but CC is much more common for written works. For example, if we publish this open access, it will be licensed with a CC license, not a software license.

It may not really matter practically if we choose MIT, BSD, or CC-BY (as the all are essentially saying the same thing), but I've rarely seen software licenses applied to academic papers. That's why I have suggested CC-BY here.

scopatz, line 24: I agree that CC is more appropriate for papers.  CC is meant for all works of art, not just software. It is common to use a CC license for prose and pretty much unheard of to use a software licence.

asmeurer, line 24: CC-BY is fine with me. It's roughly equivalent to MIT or BSD in terms of what it allows (albeit much more wordy).

scopatz, line 24: Sure, because CC-BY is designed to meet the legal requirements for its intended purpose everywhere in the world, while MIT and BSD are US-centric.

certik, line 24: Sorry for my bad grammar. I meant to write: is CC-BY an open source license? (I fixed it above.) So your question might be: "what is an open source license"? The answer is that an open source license must have a few attributes in order to be considered an open source license, they are listed here:

https://opensource.org/osd

This list was actually adapted from the Debian Free Software Guidelines (DFSG): https://www.debian.org/social_contract#guidelines, they are almost the same thing. So if a given license is not DFSG free, as determined by consensus on the [debian-legal](https://www.debian.org/legal/) mailinglist (or, for practical purposes almost the same thing, nor approved by opensource.org), then it is not an open source license.

CC-BY, that you like, was **not** an open source license, as of version 1.0 (see my link above for details). So I am strongly against using that. Version 4.0 (the version you are actually proposing) **might** be an open source license, but I wasn't able to find it clearly approved by either DFSG or opensource.org. 

@moorepants and @scopatz asked, why it matters to open source this repository? (I can't believe I have to lobby to open source things here....) But here we go: Because we, or somebody else might want to include this article perhaps with the rest of the documentation in the `sympy-doc` package, perhaps in Debian. Unless the sources of the pdf are released under an open source license, then Debian will not distribute it.

It is the same problem as with the GNU Free Documentation License (GFDL), which is not [an open source license](https://wiki.debian.org/DFSGLicenses#GNU_Free_Documentation_License_.28GFDL.29), so it is not distributed by Debian (yes, there are some finer points regarding "invariant sections", but let's not go into it here).

@scopatz raised an implied question why we would use a software license for this pdf? Because the sources are in latex, which is essentially just a software program. For example I would like to be _allowed to_ use parts of it in sympy documentation later (maybe we will, maybe we won't, but I want to have the option) --- for example the limit algorithm that I described in great detail here. If we used a license CC-BY 1.0 for this repository, we copied some stuff into sympy documentation, then Debian would have to strip it, otherwise they wouldn't be able to distribute it. What a terrible and messy thing to do by us. I am obviously strongly against it.

I know you are advocating CC-BY 4.0, but show me that it's an indisputably open source license, then I might consider it.

@scopatz mentioned that "it's pretty much unheard of to use a software licence." In fact, that is what Debian recommends since 2006, see here: https://www.debian.org/vote/2006/vote_001 where they made the exact same points I just made above, see the section "(3) Why does documentation need to be Free Software?":

>  There are a number of obvious differences between programs and documentation that often inspire people to ask "why not simply have different standards for the two?" For example, books are often written by individuals, while programs are written by teams, so proper credit for a book might be more important than proper credit for a program. 

... (read the answer at the link I provided)

Finally, Debian recommends (also on that link):

> As a consequence, other licenses should be investigated; generally it is probably simplest to choose either the GNU General Public License (for a copyleft license) or the BSD or MIT licenses (for a non-copyleft license). 

Conclusion: the general accepted practice in open source (Debian) is to use open source software licenses for both codes, as well as documentation and books.

I have indeed seen many book authors to use CC-BY 4.0, so perhaps it is an open source license. However, to make things simple, I propose we dual license this repository under the SymPy BSD license (or MIT) and CC-BY 4.0, if you think that's a better license (I am not convinced that it's a better license).

asmeurer, line 24: CC-BY isn't listed on the [OSI site](https://opensource.org/licenses/alphabetical), but interestingly enough, the contents of the site itself are licensed CC-BY 4.0.

certik, line 24: Debian webpages are dual licensed under MIT and GPL:
https://www.debian.org/license

certik, line 24: @moorepants, @scopatz, @asmeurer  --- any objections against dual licensing under MIT and CC-BY 4.0?

moorepants, line 24: That sounds fine to me.

moorepants, line 24: It will allow reuse in Debian related things and also it will allow reuse in academic publishing avenues.

scopatz, line 24: No objections from me.

scopatz, line 24: Also @certik, as a point of clarification when I said, "pretty much unheard of" I was not referring to documentation, which is often released under a software license and is generally considered part of the software project.

Rather, I was consider other large bodies of prose knowledge, such as wikipedia / wikimedia and the arxiv. Both of these are licensed under a CC license and not under MIT or similar. See section 7 of https://wikimediafoundation.org/wiki/Terms_of_Use and http://arxiv.org/help/license 

In particular, I think this paper falls into more of the arxiv category than anything else.  Since arxiv distributes the latex source and not just the resultant pdf, these source documents typically also fall under their CC or similar licence. 

I think the dual licencing is fine, but I just wanted to explain my thought process :)

certik, line 24: @scopatz Yes, you are right that the CC licenses are prevalent for such things. But based on my reasoning above, I think it's problematic. But we found a solution to dual license it, so the problem is solved.

Thanks everybody.

certik:
@moorepants would you mind fixing the two little issues:
- add MIT as a dual license next to CC-BY 4.0
- fix the gitter link

Merge with latest master to resolve issues and let's get it in.

I have another PR for the README coming, so I would like to base on top of your work once it's merged.

README.md
isuruf, line 24: FYI, CC-BY 4.0 should be compatible,
https://wiki.debian.org/DFSGLicenses#Creative_Commons_Attribution_Share-Alike_.28CC-BY-SA.29_v4.0
https://lists.debian.org/debian-legal/2015/11/msg00002.html

moorepants, line 24: For the sake of adding more to this...the same reasoning used by Ondrej for choosing MIT (i.e. if we want to be able to include this source/content in Debian related things we need MIT), we also can say "If we want to be able to include the source/content in published OA journals we need a CC license". It is problematic towards journal publishing to enforce an MIT license on journals that don't accept them.

So I think this dual license will meet all needs in this case.

Note that if we submit to SIAM (or any other closed journal) we will likely have to assign copyright to them, as they are unlikely to accept the terms of either of these licenses.

moorepants:
I've fix everything. Should be mergable.

README.md
certik, line 24: @moorepants thanks for the clarification. I agree. If we choose to submit a journal that won't accept our license, we can discuss and change the license at that point.

@isuruf good point, thanks for finding it. Yes, I posted the first link to the wiki, but CC-BY-SA is a different license to CC-BY. CC-BY is much better, as it is more permissive. However, the mailinglist thread seems to suggest that indeed CC-BY is fine with Debian. That thread is half a year old, so this shows how incredibly untested and new these licenses are.

So dual licensing seems to achieve all our objectives.

certik:
Thanks @moorepants !

README.md
asmeurer, line 24: It's fine with me too.

PR #59, opened 2016-04-19
Upload pdf

isuruf:
This will upload the pdfs to https://github.com/isuruf-bot/sympy-paper

Pull requests will only work if they are branches of https://github.com/sympy/sympy-paper. Forks will not work

Status api will give the link to the paper. (This won't work until @isuruf-bot is given permissions)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,15 @@
 sudo: required
 dist: trusty
 before_install:
-  - "sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor dvipng texlive-latex-recommended python-pygments"
+- sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended
+  texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
+  dvipng texlive-latex-recommended python-pygments
 script:
-  - make travis
-# TODO: Deploy the built file somewhere (see for instance https://harshjv.github.io/blog/setup-latex-pdf-build-using-travis-ci/)
+- make travis
+- sh upload_travis.sh
 notifications:
   email: false
+env:
+  global:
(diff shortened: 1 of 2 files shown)


certik:
This is great. How do I give it permissions? What exactly needs to be done?

isuruf:
Status api is just a convenience thing for finding the link for the paper. Otherwise, it should work after merging this

certik:
I added the `isuru-bot` in, with write access.

certik:
Let's see if it works with master. We can improve upon it as we go.

certik:
So it failed in master -- the test passes, so at least the status is OK, but it failed to upload it with the error:

```
remote: Permission to isuruf-bot/sympy-paper.git denied to isuruf-bot.
fatal: unable to access 'https://TOKEN@github.com/isuruf-bot/sympy-paper.git/': The requested URL returned error: 403
+export payload={"state":"success","target_url":"https://github.com/isuruf-bot/sympy-paper/blob/pdfs/paper-master.pdf","description":"Pdf uploading succeeded!","context":"continuous-integration/latex-pdf"}
+set +x
{
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3"
}
```

Also, it exposed your token!! So you will have to change it. I thought that Travis hides the token. We need to figure out a more secure way than this.

certik:
Because the token got exposed, I have removed write access to isuruf-bot, until this token is disabled and a new one issued.

moorepants:
This shows some builds: https://www.sharelatex.com/github/repos/sympy/sympy-paper

But the build now button didn't seem to do anything and the pdf links are dead.

isuruf:
I'll work on a branch in this repo and ping you once its working

certik:
ok!

certik:
@isuruf [here](https://github.com/espressopp/espressopp/pull/43/files#diff-354f30a63fb0907d4ad57269548329e3R23) is an example that a friend sent me. He saves a private ssh key (encrypted), and then just decrypts the ssh key before upload on Travis. Then the corresponding public ssh key must be added to the repository where you want to push. This setup is needed for automatically updating web pages, since github will not update them using the deploy keys (tokens). 

However, in our case here, the token should work, I think that's what it was designed for. I wonder if you can set it as some kind of an environment variable, so that it doesn't get exposed by git on failure, as happened above.

asmeurer:
In the past I've used `set +x` at the top of the script and `> /dev/null 2>&1` at the end of each git command to prevent the token from being leaked (but this also makes it hard to debug, since you have no output whatsoever).

PR #62, opened 2016-04-19
Add a Dockerfile to allow reproducible builds

certik:
See the comments in the file itself for instructions how to use it.


certik:
@moorepants I was able to build the article myself using this approach.

PR #38, opened 2016-04-13
First draft of physics features.

moorepants:
minted has been included for syntax highlighting (make sure to invoke the
included Makefile for the correct complilation flag)

PR #63, opened 2016-04-19
Add back a missing reference

certik:
This reference was removed in 1ea403206d9a18f76d83068e8c6f517ffea8a41f, so this commit just adds it back in.

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -164,3 +164,9 @@
   note={US Patent 7,418,369}
 }
 
+@phdthesis{gruntz1996computing,
+  title={On computing limits in a symbolic manipulation system},
+  author={Gruntz, Dominik},
+  year={1996},
+  school={Swiss Federal Institute of Technology}
+}


asmeurer:
We should restore all the removed references, perhaps to a separate bib file. I added those in there for a reason. @moorepants

moorepants:
Like I said in #38 I don't think we should add bib entries that we don't use.

The issue was that you added them all without running them through bibtex and then when I tried to get my references to work I had tons of bib entries to repair to avoid the errors. I don't really want to go through a long list of bib entries fixing all the errors if we aren't going to use them. If you'd like to go through them all and figure out the errors, then that's fine, we can add them back. But I'm not interested in fixing errors for unused refs.

certik:
We should set up bibtex to give an error when a reference is missing, and we can also set it to give an error if a reference is unused. If bibtex allows it. Then we just use Travis to take care of this.

Either way, we use the reference added in this PR, so it should be merged.

asmeurer:
In my [test merge](https://github.com/asmeurer/sympy-paper/blob/all-prs/paper.bib) branch all cictations were left intact and the paper built just fine (after a small fix regarding a duplicate citation).

PR #67, opened 2016-04-19
Use numbers instead of bullet points

certik:
Since I frequently refer to the individual bullet points, so numbers are easier to refer to.


certik:
I am just going to merge it, as I need to use it in an email.

PR #69, opened 2016-04-20
Use longtable for the features table

asmeurer:
This way, it breaks across pages. See
https://tex.stackexchange.com/questions/26462/make-a-table-span-multiple-pages.

[Build pdf](https://github.com/isuruf-bot/sympy-paper/blob/pdfs/paper-69.pdf)

### DIFF
--- a/features.tex
+++ b/features.tex
@@ -8,6 +8,7 @@ topics and application domains that SymPy services.
 
 
 \begin{longtable}[htbc]{|l|p{0.7\linewidth}|}
+\caption{SymPy Features and Descriptions}\\
 \hline
 \textbf{Feature} & \textbf{Description} \\
 \hline
@@ -63,7 +64,6 @@ Special Functions & Implements a number of well known special functions,
     Legendre polynomial, Hermite polynomial, Laguerre polynomial, and
     spherical harmonic functions.\\
 \hline
-\caption{SymPy Features and Descriptions}
 \label{features-table}
 \end{longtable}
 
--- a/features.tex
+++ b/features.tex
@@ -8,7 +8,7 @@ topics and application domains that SymPy services.
 
 
 \begin{longtable}[htbc]{|l|p{0.7\linewidth}|}
-\caption{SymPy Features and Descriptions}\\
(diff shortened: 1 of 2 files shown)


features.tex
```diff
@@ -65,8 +63,9 @@
     Legendre polynomial, Hermite polynomial, Laguerre polynomial, and
     spherical harmonic functions.\\
 \hline
-\end{tabular}
-\end{table}
+\caption{SymPy Features and Descriptions}
+\label{features-table}
```
scopatz: Table captions typically appear on the top of tables, not on the bottom, right?

scopatz:
Looks good, except for the caption location.

features.tex
asmeurer: I can't figure out how to move it.

scopatz: Oh weird ok.  There is probably some longtable option.

scopatz: Ok nm. it looks like it will render correctly

asmeurer: It is under the table this way (previously it was over).

scopatz: Hmm Section 3 of the docs probably explain how to change this: http://ctan.mirrors.hoobly.com/macros/latex/required/tools/longtable.pdf

Have to run to class right now, though.

asmeurer: Figured it out (you have to put `\\` at the end of the line).

asmeurer:
There is an issue here. The bottom of the table has some extra lines sticking out of it.
screen shot 2016-04-20 at 12 59 35 pm

Any idea what's causing that?

certik:
Regarding the extra lines, perhaps try to move the `\label` line from the bottom to the top of the table or something like that.

asmeurer:
Good catch. I didn't notice that the label should go inside the caption. Should be ready to go now.

asmeurer:
To whoever included booktabs (@moorepants?), should we reformat the table to use that?

moorepants:
I think so. The booktabs look tons nicer.

asmeurer:
Going to merge this now to keep the PR count down. But let's definitely figure out how to format the tables.

PR #73, opened 2016-04-20
Add authors.json and related files

certik:
The `authors.json` contains all the information about authors of this article. Use the provides scripts to parse this information into the form that we need. E.g. to print information about each author into the terminal, do:

```
$ ./list_text.py 
List of authors:

Author: Ondřej Čertík
Affiliation: Los Alamos National Laboratory
Address: P.O. Box 1663, Los Alamos, New Mexico 87545, USA
GitHub profile: https://github.com/certik
Article commits: https://github.com/sympy/sympy-paper/commits?author=certik
Article comments on issues and PRs: https://github.com/sympy/sympy-paper/search?q=certik&type=Issues

Author: Isuru Fernando
Affiliation: University of Moratuwa
Address: Bandaranayake Mawatha, Katubedda, Moratuwa 10400, Sri Lanka
GitHub profile: https://github.com/isuruf
Article commits: https://github.com/sympy/sympy-paper/commits?author=isuruf
Article comments on issues and PRs: https://github.com/sympy/sympy-paper/search?q=isuruf&type=Issues
```

To get the list of authors in Latex so that we can paste it into the main latex file, do:

```
$ ./list_latex.py 
List of authors:

\author{Ondřej Čertík}

\author{Isuru Fernando}
```

(More information should be exported to Latex, we can improve the script later.)

/cc @asmeurer, @scopatz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 all:
+	cd authors; ./list_latex.py
 	pdflatex -shell-escape paper.tex
 	bibtex paper.aux
 	pdflatex -shell-escape paper.tex
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ all:
 	pdflatex -shell-escape paper.tex
 	pdflatex -shell-escape paper.tex
 travis:
+	cd authors; ./list_latex.py
 	pdflatex -shell-escape --halt-on-error paper.tex
 	bibtex paper.aux
 	pdflatex -shell-escape --halt-on-error paper.tex
--- a/authors/README.md
+++ b/authors/README.md
@@ -1,0 +1,17 @@
+# Authors List
+
+## Usage
+
+* Call `./prettify.sh` after modyfing `authors.json`.
+* Use `./list_text.sh` to list the authors as text
+* Use `./list_latex.sh` to list the authors as latex source code
+
+## How to add a new author
+
+Send a PR adding a new author into the json file. This PR must be approved by
+Anthony Scopatz. We follow the criteria stated in the main
+[README file](https://github.com/sympy/sympy-paper/blob/master/README.md#authorship-criteria).
+Anthony will make sure that the new author meets all the criteria. For
+clear-cut cases (when nobody has any doubts that the criteria have been met),
+Anthony will simply merge the PR. For borderline cases, Anthony will consult
+the community.
--- a/authors/authors.json
+++ b/authors/authors.json
@@ -1,0 +1,16 @@
+[
+    {
+        "email": "ondrej.certik@gmail.com",
+        "github_id": "certik",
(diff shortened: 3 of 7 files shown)


asmeurer:
Could you make it write a LaTeX file (which would be committed to the repo), so we can just \include that file in the paper.tex?

certik:
@asmeurer sure, I've implemented it in 6817ac393c3fc21858a963ae226ca111b8af95ea.

All the scripts work in both Python 2 and 3. The `Makefile` now generates `authors.tex` which is then included from the main document. UTF-8 works up to this point. The main document doesn't work with UTF-8, and so my name is missing some characters. That can be fixed by adding:

```
\usepackage[utf8]{inputenc}
```

but then some verbatim environments fail, so I will tackle this in a separate PR, as it is not related to authors.

moorepants:
If you want full encoding support we can compile with xelatex or luatex. I think that is the most common thing to do now. I'm not sure which one is preferred.

certik:
@moorepants I am not sure either (in the past `\usepackage[utf8]{inputenc}` worked for me with just `pdflatex`). But let's tackle that in a separate PR, I want to keep the change introduced in this PR to be simple.

scopatz:
LGTM

asmeurer:
Any reason not to use `Ond\v{r}ej \v{C}ert\'{\i}k` for your name? 

Anyway, if we ever add Unicode pretty printing examples to the paper (seems likely), we'll need proper UTF-8 support.

certik:
@asmeurer yes, if I want to print to console, or copy & paste into the online journal application (depends on the journal). That's why UTF-8 was invented, so we should just use it.

certik:
Travis passes, looks like there are no more objections, let's merge it and improve upon it as we go.

ashutoshsaboo:
This seems to be pretty nice @certik - especially the latex printing part. 

Just one question. How are we planning to add the authors? Will you be adding the authors yourself @certik , or does everyone need to push PRs for the same? @asmeurer @moorepants

certik:
It's described in the Readme in the authors directory. Please send PRs for
people that you think should be there, but do not merge it.

ashutoshsaboo:
@certik Ohh sure! Sorry didn't look into it. :+1:

certik:
No problem. In the first iteration let's put there people where nobody has
doubts should be there, easily satisfying the 1-4 authorship criteria, so
that Anthony can merge it without issues. We'll handle the bordeline cases
at the end, if there are any.

aktech:
This is how it looks now:

os

certik:
Yep, we have to fix the unicode. We should now take the SIAM journal latex
template, and use their commands to typeset the authors properly.

PR #75, opened 2016-04-20
Allow all UTF-8 characters

Upabjojr:
This should allow UTF-8 characters (like those  @certik 's name) to be input.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -1,4 +1,5 @@
 \documentclass{article}
+\usepackage[utf8]{inputenc}
 \usepackage{authblk}
 \usepackage{hyperref}
 \usepackage{graphicx}


certik:
This PR gives me the following error locally:

```
Overfull \hbox (69.99886pt too wide) in paragraph at lines 96--100
\OT1/cmr/m/n/10 pres-sion through the log-i-cal con-nec-tor class, in this case
 []\OT1/cmtt/m/n/10 And(Q.positive(x), Q.positive(y))\OT1/cmr/m/n/10 . 
[13]

! Package inputenc Error: Unicode char \u8:─ not set up for use with LaTeX.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              

l.128 \end{verbatim}

```

Upabjojr:
LaTeX can be really complicated sometimes.

moorepants:
I'm pretty sure a better solution is to compile with xelatex or luatex, then utf symbols can be typed directly into the paper source. To fix that, simply change the travis install instruction to grab one of those packages and then change the make file from pdflatex to which every one you choose.

certik:
But do journals allow xelatex?

moorepants:
Journals only accept the pdf. They generally don't care how you made it. The sad thing is that they often convert your pdf to word and then publish!

asmeurer:
According to https://www.siam.org/journals/sisc/authors.php, for SIAM authors submit TeX files.  It probably differs from journal to journal.

certik:
@moorepants yes, as @asmeurer said SIAM accepts tex, and other journals I've used before accept tex. So they need to be able to build it, and as such we must be careful about these things and make sure things can be easily built (also using too many hard to install packages might break it).

moorepants:
I personally think that this journal's issue, not ours. If we follow their rules and submit what they ask, then it is up to them to deal with it after then.

certik:
@moorepants it's not just that. It's also other places, like Arxiv, which only seems to support pdflatex: https://arxiv.org/help/submit_tex. We decided to use PeerJ preprint, which only accepts pdf (https://peerj.com/about/preprints/scope-and-instructions/), so that's easier.

I've only used pdflatex and I use unicode, it works just fine. I'll work on the verbatim part. I think it's coming from some SymPy output ---- honestly, why don't we just use latex output from sympy? The unicode output is only a workaround for a terminal, to look a bit better.

asmeurer:
I guess the question is, do we want to show an example of Unicode pretty printing from SymPy? If so, then we should figure out how to make it work. If not, then we can work around it. I think we might like to (printing is a relatively important feature).

certik:
I looked at how to workaround it and still use pdflatex. Here are a few tips:

http://www.codeotaku.com/journal/2010-03/latex-symbols-in-verbatim/index
http://tex.stackexchange.com/questions/40014/encoding-problem-with-verbatiminput-package-inputenc-error-unicode-char-u8

The only thing I have to figure out is how is it possible that it works when the package `\usepackage[utf8]{inputenc}` is not loaded, but fails when it is. That's something I don't understand. Otherwise it seems it's just a matter of defining the mapping from the unicode symbols to a latex representation of the same symbol. Latex is very rich, it has tons of symbols, so we should be able to do this mapping. Otherwise (or in addition to), if we want to show how it actually looks in a terminal, then we can simply post an image.

certik:
Anyway, closing this in favor of #88, which fixes it.

PR #78, opened 2016-04-21
Add SIAM style files

asmeurer:
There is some build error, but I'm not sure what's going on.

```
! Illegal parameter number in definition of \@tempc.
<to be read again>
                   1
l.42 \maketitle

?
```

If you know how to fix it, feel free to push to this branch.

### DIFF
--- a/authors/list_latex.py
+++ b/authors/list_latex.py
@@ -9,5 +9,5 @@ author_list = load(open("authors.json"))
 with open("../authors.tex", "wb") as f:
     for author in author_list:
         f.write(("\\author{%s}\n" % author["name"]).encode("utf-8"))
-        f.write(("\\affil{%s}\n" % author["institution"]).encode("utf-8"))
+#        f.write(("\\affil{%s}\n" % author["institution"]).encode("utf-8"))
         f.write("\n".encode("utf-8"))
--- a/authors/list_latex.py
+++ b/authors/list_latex.py
@@ -3,15 +3,16 @@
 from __future__ import print_function
 
 from json import load
+from io import open
 
 author_list = load(open("authors.json"))
 
-with open("../authors.tex", "wb") as f:
-    f.write("\\author{%\n")
+with open("../authors.tex", "w") as f:
+    f.write(u"\\author{%\n")
     for n, author in enumerate(author_list):
-        f.write(("%s%%\n" % author["name"]).encode("utf-8"))
-        f.write(("\\thanks{%s (\\email{%s}).}\n" \
-                % (author["institution"], author["email"])).encode("utf-8"))
+        f.write((u"%s%%\n" % author["name"]))
+        f.write(u"\\thanks{%s (\\email{%s}).}\n" \
+                % (author["institution"], author["email"]))
         if n < len(author_list) - 1:
-            f.write("\\and\n".encode("utf-8"))
-    f.write("}\n")
+            f.write(u"\\and\n")
+    f.write(u"}\n")
--- a/authors/list_latex.py
+++ b/authors/list_latex.py
@@ -7,7 +7,11 @@ from json import load
 author_list = load(open("authors.json"))
 
 with open("../authors.tex", "wb") as f:
-    for author in author_list:
-        f.write(("\\author{%s}\n" % author["name"]).encode("utf-8"))
-#        f.write(("\\affil{%s}\n" % author["institution"]).encode("utf-8"))
-        f.write("\n".encode("utf-8"))
+    f.write("\\author{%\n")
+    for n, author in enumerate(author_list):
+        f.write(("%s%%\n" % author["name"]).encode("utf-8"))
+        f.write(("\\thanks{%s (\\email{%s}).}\n" \
+                % (author["institution"], author["email"])).encode("utf-8"))
+        if n < len(author_list) - 1:
+            f.write("\\and\n".encode("utf-8"))
+    f.write("}\n")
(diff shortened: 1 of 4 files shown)


asmeurer:
It has something to do with the authors file. @certik

asmeurer:
I think their style file is incompatible with `authblk`. Even if I remove the `\input{authors}`, but leave the `\usepackage{authblk}` it fails. So we need to rewrite the authors stuff to be compatible with their thing.

asmeurer:
I removed the authors in a commit so that it will build and you can see what it looks like.

certik:
Yes, authblk is needed for the authors to work. Don't merge this PR yet. Let's figure out the syntax to get authors working. I can help.

certik:
It looks like the syntax should be like this:

```
\author{Dianne Doe%
  \thanks{Imagination Corp., Chicago, IL (\email{ddoe@imag.com}).}%
  \and
  Paul T. Frank%
  \thanks{Department of Applied Math, Fictional University, Boise, ID
    (\email{ptfrank@fictional.edu}, \email{jesmith@fictional.edu}).}
  \and
  Jane E. Smith%
  \footnotemark[3]
}
```

asmeurer:
[Here's](https://github.com/isuruf-bot/sympy-paper/blob/pdfs/paper-78.pdf) a build of the paper.

certik:
@asmeurer, I just pushed in a fix in 04556ee14db11de9e42487c4032e179b63d700b8.

asmeurer:
@isuruf's build doesn't seem to be updated, but for me, this build looks like this
screen shot 2016-04-21 at 12 50 05 pm

So it only shows the one author.

certik:
Yes, see the syntax I posted above vs what's in authors.tex. I am fixing it now.

certik:
@asmeurer it should be fixed now.

authors/list_latex.py
```diff
@@ -7,7 +7,11 @@
 author_list = load(open("authors.json"))
 
 with open("../authors.tex", "wb") as f:
```
asmeurer: Why do you open this file as binary? This is making this not work in Python 3.

certik: Actually, this is to make it work in Python 3. It must have been broken by my recent fix, as the file in master works in both Python 2 and Python 3, I tested it. The file in this PR is only tested with Python 2, as I don't have Python 3 here.

Would you mind fixing it for Python 3, and also making sure it works in Python 2?

certik:
@asmeurer, here is the latest generated pdf: https://github.com/isuruf-bot/sympy-paper/blob/pdfs/paper-siam.pdf

asmeurer:
Unicode still doesn't work, but it looks good. Go ahead and merge if you see no other issues.

authors/list_latex.py
```diff
@@ -3,15 +3,16 @@
...
-        f.write(("\\thanks{%s (\\email{%s}).}\n" \
-                % (author["institution"], author["email"])).encode("utf-8"))
+        f.write((u"%s%%\n" % author["name"]))
+        f.write(u"\\thanks{%s (\\email{%s}).}\n" \
+                % (author["institution"], author["email"]))
         if n < len(author_list) - 1:
-            f.write("\\and\n".encode("utf-8"))
-    f.write("}\n")
+            f.write(u"\\and\n")
+    f.write(u"}\n")
```
certik, line 18: Thanks for fixing it, it now looks much better, and I tested in Python 2 and it works. I couldn't get it to work before.

certik:
@asmeurer how do you comment per commit on github PR? Previously you could click on the commit, and then comment at the bottom. Now the option seems removed, as there is no comment box below a commit. So I commented on the last line of that commit above.

certik:
I think it looks good. Merging.

authors/list_latex.py
asmeurer: Yeah, from what I remember, `io.open` + unicode literals is the ticket for writing Unicode files in Python 2 (of course it just works in Python 3). Writing a binary file could have also worked, but it's way more messy, because you have to do all the encoding yourself.

asmeurer:
I don't know. GitHub recently streamlined the interface so that commits are tied back to the PR. I guess they removed the whole-commit commenting when they did that.

authors/list_latex.py
certik: Just so you know, your change actually did break it in Python 2 inside docker:

```
Traceback (most recent call last):
  File "./list_latex.py", line 13, in <module>
    f.write((u"%s%%\n" % author["name"]))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0159' in position 3: ordinal not in range(128)
```

But it works with Python 2 on my desktop. The fix seems to require to add the following to `/usr/lib/python2.7/sitecustomize.py`:

```
import sys
reload(sys)
sys.setdefaultencoding('utf8')
```

So Python 2 unicode support seems to be seriously screwed up.

asmeurer: Oh maybe you have to use `io.open(encoding='utf-8')`. Does that fix it?

certik: yes, it does!! Thanks a lot. Much better than messing with /usr/lib/python2.7/sitecustomize.py

asmeurer: That's my bad. I forgot that critical detail.

certik: I just submitted #83 that fixes it.

certik: @asmeurer what's the underlying cause of this? Python 2 can sometimes be configured to use utf8 by default, and sometimes not, depending on a local user configuration? And so in order to be portable, one must always set this inside the `open` statement?

asmeurer: I think that's right, but I'm not 100% sure of it.

PR #86, opened 2016-04-21
Disable the push of pdf for now

certik:
This is due to failures like this one:

https://travis-ci.org/sympy/sympy-paper/builds/124812679


certik:
@isuruf we need to make the upload script a lot more robust. I created an issue #87 for it.

PR #88, opened 2016-04-22
Enable Unicode Support (Alternative I)

certik:
Note: there is an alternative patch at #89 that might be better.


certik:
Note that this actually fixes two bugs, as previously, the unicode characters for "-" were simply skipped in the pdf and nobody noticed.... Now they are visible.

certik:
@asmeurer, please merge this PR. It fixes 3 big bugs, and doesn't introduce a single new one.

certik:
Thank you.

Issue #87, opened 2016-04-21
Fix the upload script


certik:
I had to disable pdf upload in #86, as it was not robust. To make it robust, we should:
- [ ] enable `set -e`, i.e. fail on any error in the upload script
- [x] disable notifications for now (I think they already are), since they don't work
- [x] figure out a robust way to do `git push`. Perhaps it could work to pull first and then push. It depends if the failure is caused by some other Travis test pushing in changes after we pulled but before we pushed. Or if the failure is caused by something else, we need to investigate and fix it, perhaps try a few times, etc.

moorepants:
Would this be of interest: https://github.com/ewanmellor/gh-publisher/ ?

asmeurer:
If we're not pulling before we push we should definitely do that.

isuruf:
Failures mentioned were because of PRs coming from forks. We should exit early if `GH_TOKEN` is empty

asmeurer:
Is this fixed?

certik:
The script is still not using `set -e`, so this still needs to be fixed. The other two points seem to be fixed.

PR #90, opened 2016-04-22
Fix for #87

isuruf:
- Exit early if PR is from fork
- Retry git push several times

### DIFF
--- a/upload_travis.sh
+++ b/upload_travis.sh
@@ -1,3 +1,10 @@
+set +x
+if [ "${GH_TOKEN}" = "" ]; then
+    echo "PDF not uploading because this is a PR from a fork"
+    exit 0
+fi
+set -x
+
 export commit_id=`git log -1 --pretty=format:"%H"`
 
 git config --global push.default simple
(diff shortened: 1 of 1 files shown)


certik:
Looks like it works, let's try it out.

asmeurer:
Let's also remind everyone to push up branches to origin, not to forks. Everyone should have push access. If someone wants to contribute and doesn't they can be added.

certik:
@asmeurer is the motivation to get the pdf pushed automatically from the PR?

asmeurer:
Yes, @isuruf's bot won't build the PDF unless the branch is on sympy/sympy-paper. It's obviously not a requirement to have it, since Travis itself will build the PDF either way, but it's nice for the reviewers to have. 

Also, as we move into the editing stage, I think it will be easier if we push to shared branches, instead of everyone having their own branches with tons of merge conflicts among each other (but I'm still unclear on what exactly the best workflow for this will be).

certik:
I see. Makes sense.

PR #91, opened 2016-04-25
Remove Wikipedia references

asmeurer:
They were causing the build to fail, and I don't think we should be
referencing Wikipedia anyway.

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -3317,22 +3317,6 @@ ng {C}omputations in {C}ommutative {A}lgebra},
   url = {https://reference.wolfram.com/language/tutorial/FlatAndOrderlessFunctions.html},
 }
 
-@misc{wiki:fourier,
-   author = "Wikipedia",
-   title = "Fourier series --- Wikipedia{,} The Free Encyclopedia",
-   year = "2016",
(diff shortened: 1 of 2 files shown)


certik:
What was the build failure?

asmeurer:
Same failure reported at https://github.com/sympy/sympy-paper/issues/93

certik:
Ah ok.

asmeurer:
I guess the url formatting was wrong (other refs in the bib file don't use `\url`). Even so, I don't think we should be referencing Wikipedia, so I just removed them.

leosartaj:
This is what it says if we reference a [page](https://en.wikipedia.org/w/index.php?title=Special:CiteThisPage&page=Formal_power_series&id=714764056) on wikipedia.

> When using the LaTeX package url (\usepackage{url} somewhere in the preamble), which tends to give much more nicely formatted web addresses, the following may be preferred:
> 
> @misc{ wiki:xxx,
>    author = "Wikipedia",
>    title = "Formal power series --- Wikipedia{,} The Free Encyclopedia",
>    year = "2016",
>    url = "\url{https://en.wikipedia.org/w/index.php?title=Formal_power_series&oldid=714764056}",
>    note = "[Online; accessed 26-April-2016]"
>  }

Also, do you have anymore references in mind for Formal Power Series and Foruier Series?

PR #99, opened 2016-05-02
Use Pythontex

isuruf:
This embeds python sessions with output. Possible fix for #9, only problem I see is that the output is not tested.

We can also add pythontex to projects that use SymPy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ before_install:
 - sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended
   texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
   dvipng texlive-latex-recommended python-pygments texlive-extra-utils pgf
+- pip install pygments sympy --user
 script:
 - make travis
 - sh upload_travis.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 before_install:
 - sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended
   texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
-  dvipng texlive-latex-recommended python-pygments
+  dvipng texlive-latex-recommended python-pygments texlive-extra-utils
 script:
 - make travis
 - sh upload_travis.sh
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -3,7 +3,8 @@ dist: trusty
 before_install:
 - sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended
   texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
-  dvipng texlive-latex-recommended python-pygments
+  dvipng texlive-latex-recommended python-pygments texlive-extra-utils pgf
+- pip install pygments sympy --user
```
asmeurer: sympy==1.0 (the paper should always build against 1.0, even in the future)

certik:
I like this. Though I would rather see a "doctesting" style, i.e. have both input and output in the tex itself (that way, it's just a latex document and journals will have no problems with it), and just use Travis to doctest it, to ensure that the output is correct.

The current solution doesn't show the output. Last thing we want is the journal to run this themselves with an incorrect version of sympy, or some other broken configuration.

So I am -1 on the current PR as it is.

asmeurer:
How can we doctest the code? I like that idea, because that means that we can just write the TeX directly, and the testing that it is correct is done completely separately. So there is no need for more dependencies, which run the risk of the journal rendering the paper incorrectly.

isuruf:
We can have a python script go through the text files looking at blocks like

```
%%doctest
\begin{verbatim}
>>> from sympy import *
>>> x, y = symbols('x y')
>>> expr = x*y + 2
\end{verbatim}
```

It will then generate a string using all the commands and then doctest it.

What do you think?

asmeurer:
Any reason to not just look for `\begin{verbatim}`. Are there verbatim blocks that shouldn't be doctested?

isuruf:
`verbatim` blocks can have other stuff like Mathematica code.

asmeurer:
Let's make verbatim doctest by default and have a `%%nodoctest` marker for non-Python blocks.

certik:
Alternatively, we can simply search for `>>>`, which I think is exactly what doctest does. However, we need to use Latex (`%`) comments instead of Python comments (`#`) for doctest directives like "skip doctest" (for Mathematica prompts).

asmeurer:
I would also like to "doctest" the `gamma` example (at least check that it is valid Python). Maybe we should add some example usage after to show how it works.

asmeurer:
(that's page 4 [here](https://github.com/isuruf-bot/sympy-paper/blob/fd1b03603fe6f5f9f9be6b905de6de68a8d11942/paper-99.pdf))

asmeurer:
I agree with Ondrej here. It's better to include the output in the tex file, but check it, than to generate it automatically.

PR #101, opened 2016-05-04
Cleanups of the paper

asmeurer:
I've added a TODO for items that I see that need to be done. I'd like to merge this soon, so that we can continue work.


certik:
+1 to merge as is.

I like that you keep the supplement in the same pdf, for easier manipulation. Before submitting, we'll probably have to split it.

Let's add more authors, following the guidelines here: https://github.com/sympy/sympy-paper/blob/master/authors/README.md

certik:
Ok, I've added the authors in #102.

asmeurer:
Thanks Ondrej. Will you be able to help with any of the items in the [TODO](https://github.com/sympy/sympy-paper/blob/trimming/TODO.md)? I talked with @scopatz and we agreed that the most important things right now are to get an introduction and conclusion, so that there are no major missing sections. Then everyone can go in and review the paper as a whole.  But there are also several smaller things that need to be fixed as well, as well as several sections that need cleaning up.

certik:
Yes, I'll work on the quantum section, I'll polish up the limits a bit more. I want to add more citations etc. I want to help with the introduction and conclusion as well. There is plenty of work to do.

asmeurer:
How early can you commit to doing these things? I want to start doing mass reviews next week, meaning all major content should be included at that point. 

By the way, for the Gruntz section, I think the details on the algorithm should be moved to the supplement. The algorithm itself isn't novel to SymPy, and we don't go into that much detail for any of the other algorithms in SymPy. It is a nice thing to have in the supplement, though (and either way, it needs to be written up as actual text).

certik:
I'll do my best over the weekend.

PR #105, opened 2016-05-05
Replace Example's Input/Output with >>>

aktech:


moorepants:
If that's all of them, +1.

aktech:
> If that's all of them, +1.

Yeah, I had grep it, couldn't find more.

PR #116, opened 2016-05-11
More Fixes

asmeurer:


paper.tex
```diff
@@ -35,6 +35,8 @@
 
 \usepackage{longtable}
 
+\hyphenation{Math-e-ma-ti-ca pol-y-morph-ism}
```
certik: That's not the hyphenation that I got here: https://www.hyphenation24.com/?term=Mathematica, even though it said it can't find the word, so it might be inaccurate.

As to `polymorphism`, again, Webster http://www.merriam-webster.com/dictionary/polymorphism suggests poly·mor·phism, which is different to what you put there.

asmeurer: Yeah, that's why I added it. LaTeX thinks it's `Math-e-mat-ica`. I guess it depends on how you pronounce it. 

For Webster, I guess I'll use what they have. That's different from the pronounciation (you can see they have \ˌpä-lē-ˈmȯr-ˌfi-zəm).  I've always thought of it as morph-ism, not mor-phism. 

Anyway, in both cases, that's the default hyphenation that LaTeX uses, and it causes overfull \hboxes. Maybe we should just rewrite the sentence.

certik: I would worry about overflows once the article is written, as you mentioned in #117.

asmeurer: Yeah, I just saw this at the end of the build log and wanted to see if I could fix it. But let's not worry about it now.

PR #119, opened 2016-05-11
Use T1 font encoding and Latin Modern fonts

certik:
This makes sure that unicode characters like Č are present in the font, and
thus displayed properly.

Before this PR:
old
after this PR:
new

Notice the position of the `v` accent above `C`. Before it was positioned incorrectly, because Latex doesn't have the Č character in the font, and so it simply puts the accent in the middle of the letter, since that's the best it can do --- and so the result looks extremely crappy (perhaps not if you are not used to seeing such letters, but it is obvious to me from first look that the accent is positioned incorrectly). However, the proper way is to have all accented letters in the font itself, since the positioning of accents is specific to each letter.

However, this PR introduces another, unrelated, problem (also apparent from the screenshots above) --- before the font seems to be fully vectorized and looks good if I enlarge the pdf, while in this PR, the font seems rasterized, and so it looks crappy when enlarged. Currently I don't know what is causing it, nor how to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 before_install:
 - sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended
   texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
-  dvipng texlive-latex-recommended python-pygments
+  dvipng texlive-latex-recommended python-pygments lmodern
 script:
 - make travis
 - sh upload_travis.sh
--- a/paper.tex
+++ b/paper.tex
@@ -1,6 +1,7 @@
 % SIAM Article Template
 \documentclass[review]{siamart0216}
 
+\usepackage[T1]{fontenc}
 \usepackage[utf8]{inputenc}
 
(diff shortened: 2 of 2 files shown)


asmeurer:
Looks like this to me 
screen shot 2016-05-11 at 4 31 49 pm

asmeurer:
Also the I is different. Assumedly it is also more correct now.

certik:
I implemented a solution, that also works for me, in 0ebfe30f90718937e0039e78e3610b6107fa8121 (see the commit log for detailed explanation). Now it looks like this for me:

new2

Which is really good (they updated the position of the accent in `Č`, I like the new version better). As far as `Í`, the accent depends on the font used.

The journal might want to use their own fonts. But I think the current solution is good for now, at least our manuscript will be correct.

asmeurer:
We are using the Journal style files, so this should be their own fonts.

certik:
@asmeurer they don't seem to specify fonts, as far as I know. Either way, I submitted #121 with just the T1 fix, without the font change.

certik:
Ok, #121 is in, and I created #123 for the Latin Modern fonts. I am closing this PR.

PR #120, opened 2016-05-11
Fix a bug in the SIAM Article Template

### LINKED ISSUE #96: Ondřej's name renders as ONDřEJ ČERTÍK
This might just be a bug with the SIAM template, which converts the author names to uppercase. If you use `On\v{r}ej` it gives `ONDŘEJ` correctly. But right now it's using the Unicode characters and the `Ř` is kept lowercase. 

@certik

certik:
The problem is that the SIAM Article Template is using \uppercase instead of
\MakeUppercase, as discovered by:

http://tex.stackexchange.com/questions/289465/stylesheet-does-not-capitalize-polish-letters-correctly

We fix it by modifying one of the commands from the template. We will have to
bring this up with the journal while submitting.

Fixes #96.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -1,6 +1,10 @@
 % SIAM Article Template
 \documentclass[review]{siamart0216}
 
+% Apply a fix to the SIAM Article Template
+\input{siamart0216_uppercase_fix}
+
+
 \usepackage[utf8]{inputenc}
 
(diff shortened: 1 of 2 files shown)


certik:
@asmeurer, would you mind merging this one? It fixes a bug.

PR #121, opened 2016-05-11
Use T1 font encoding

certik:
This makes sure that unicode characters like Č are present in the font, and
thus displayed properly.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -1,6 +1,7 @@
 % SIAM Article Template
 \documentclass[review]{siamart0216}
 
+\usepackage[T1]{fontenc}
 \usepackage[utf8]{inputenc}
 
 \usepackage{hyperref}


asmeurer:
I don't suppose we can use similar tricks to make the Unicode pretty printing work.

certik:
I think we might, I haven't even looked at that yet. Especially with the Latin Modern fonts, perhaps they contain the unicode characters that we need.

Right now let's fix these small issues. If you are ok with this fix, then please merge it, so that I can move on to the bigger unicode issues.

certik:
@asmeurer can you merge this if you are ok with it?

asmeurer:
Well if you want to check, this is the output I want to show

```
>>> pprint(Integral(sqrt(phi0 + 1), phi0))
⌠
⎮   ________
⎮ ╲╱ φ₀ + 1  d(φ₀)
⌡
```

certik:
Thanks for merging this. I created #122 for the output you want to show. I am looking into that now.

PR #123, opened 2016-05-12
Use Latin Modern fonts

certik:
This resolves the bitmap fonts issue introduced in
6e5dda49a9bf0c2ebacaa9f3e0d45e4982bd9d35. The problem is that the standard
Computer Modern fonts are in OT1 encoding, and so when the 6e5dda4 commit
requested T1 font, the bitmap fonts were used, which look ugly in high
resolution.

One solution, implemented in this commit, is to use the Latin Modern fonts,
e.g. in Ubuntu:

```
apt-get install lmodern
```

Which support T1. More information at:

http://tex.stackexchange.com/questions/1291/why-are-bitmap-fonts-used-automatically

The other solution is to use `cm-super`, but as explained at:

http://tex.stackexchange.com/questions/1390/latin-modern-vs-cm-super

it seems that Latin Modern is the way to go to obtain high quality fonts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 before_install:
 - sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended
   texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
-  dvipng texlive-latex-recommended python-pygments
+  dvipng texlive-latex-recommended python-pygments lmodern
 script:
 - make travis
 - sh upload_travis.sh
--- a/paper.tex
+++ b/paper.tex
@@ -1,6 +1,7 @@
 % SIAM Article Template
 \documentclass[review]{siamart0216}
 
+\usepackage{lmodern}
 \usepackage[T1]{fontenc}
 \usepackage[utf8]{inputenc}
 


asmeurer:
The differences here are extremely subtle. It seems to only affect the accents on your name, the thickness of the `[]` characters used for references and the `+` character in text (in `C++`) (`[]` is thicker and `+` is thinner in master), the height of the `-`, `_`, `^`, and `~` characters in code blocks, and some minor kerning differences randomly across the text (which spill into a different word wrapping in a few places, making it a little difficult to compare parts of the document). 

I have a hard time judging which is better. The `_` looks better in master (it's too low in this PR), and the `~` looks better in this PR (it's centered, vs. too high in master). I've no opinion on which is better on any of the other differences.

asmeurer:
Also, the characters in your name don't appear bitmapped in either case.

certik:
The subtle changes, as you reported, are minor, I am fine either way, that is not a problem.

The problem is that the whole document is bitmapped, on linux (Travis), while I assume that on your Mac, it is not bitmapped. Compare e.g. before:

https://github.com/isuruf-bot/sympy-paper/raw/7adaedfbc838535e717bdef42684d311bc45439a/paper-121.pdf

and after:

https://github.com/isuruf-bot/sympy-paper/raw/326d5152bd5454d3f306d62a04157c772c788cd0/paper-123.pdf

That is the main issue.

asmeurer:
Oh wow. Yeah, the master pdf definitely doesn't look like that for me. Travis must have a crappy bitmap version of whatever font is being used. If this fixes that issue, I'm fine with it.

certik:
@asmeurer not just Travis, but also all my linux computers that I tried (Ubuntu 14.04, Ubuntu 16.04). Also the links I posted in the commit log suggest that this is the default behavior to use bitmap fonts. Using Latin Modern indeed fixes it for both me and you, and Travis.

certik:
@asmeurer if you are ok with that, can you merge this?

certik:
Thank you.

PR #127, opened 2016-05-13
Add some citations

asmeurer:
Just MathJax for now. It's the first result from Google Scholar, so I hope it's the right one to use. 

I might go in and add others later. Feel free to push to this branch, or to merge it. 

@moorepants how should we cite PyDy (in the introduction)?


moorepants:
PyDy citation: https://github.com/pydy/pydy#citation

PR #128, opened 2016-05-13
Unicode pprint using xetex pdf figure

asmeurer:
Working on the Unicode pprint example, using xetex to generate a pdf "figure".

You can see the basic idea of how it works. Some issues:
- [x] The example is indented
- [x] There is an extra line below the text
- [x] The font size doesn't exactly match
- [ ] I am using DejaVu Sans Mono because that has all the characters (hopefully this will build on Travis). But this looks different from all the other examples. We might need to find a more standard LaTeX font that has the characters, or else set the mono font for the whole paper. 
- [x] The spacing between the integral sign characters is not right.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,11 +11,10 @@
 ## Intermediate documents:
 *.dvi
 *-converted-to.*
-paper.pdf
 # these rules might exclude image files for figures etc.
 # *.ps
 # *.eps
-# *.pdf
+*.pdf
 
 ## Bibliography auxiliary files (bibtex/biblatex/biber):
 *.bbl
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,10 @@ before_install:
   texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
   dvipng texlive-latex-recommended python-pygments lmodern
 script:
-- make travis
+- make
 - sh upload_travis.sh
 notifications:
   email: false
 env:
   global:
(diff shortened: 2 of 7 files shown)


asmeurer:
Here's how it looks, in case Travis doesn't build.

screenshot 2016-05-13 13 45 06

asmeurer:
Another issue: even though we are using DejaVu Sans Mono, the integral sign characters do not line up (that's the whole point of using that font, because it's the only one that lines up those characters correctly). This is how it looks in my terminal

screenshot 2016-05-13 13 49 57

asmeurer:
Apparently you can just `\usepackage{DejaVuSansMono}` to use DejaVu Sans Mono (there are [other](http://www.tug.dk/FontCatalogue/typewriterfonts.html) fonts as well). The default font size for it is too big, though.

asmeurer:
I think the spacing issue is because it is using an old version of the font, which doesn't have have the longer integral extension character (this is a common issue with DejaVu Sans Mono actually). I'm unsure, though, since XeTeX is supposed to be getting the font from the system (my system definitely has the newer font), but it must be defaulting to the one installed by LaTeX first.

asmeurer:
So I guess the spacing was coming from LaTeX. Still, I'm concerned that an older version of the font may break things. Let's see what the Travis build looks like. If it breaks, we can include a local copy of the newest version of the font, so that it always builds correctly.

asmeurer:
This is what it looks like now

screenshot 2016-05-13 15 46 28

asmeurer:
@certik can you build this branch from the Docker container, and show me what the output looks like (the top of page 12)?

asmeurer:
From my point of view, this is done. There are still two questions, though:
- [ ] Will older versions of DejaVu Sans Mono break the rendering of the integral sign? If so, the fix is easy.
- [ ] Will the journal be OK with using DejaVu Sans Mono as our monospace font, instead of the default one? 
  
  If they aren't, this will take more work. The default monospace font (I'm not sure what it's called) doesn't support any of the Unicode characters we need. We would need to substitute them in, hopefully from a font where it looks correct with the default font. That would be more work, but I think it's doable. My main concern is that if we have to do that, that the different fonts won't line up correctly. We would have to figure out how to make LaTeX render each font character exactly the same width. 
  
  And anyway, it would be a "faked" example. My preference is the way we have it now. A third option, as I've mentioned elsewhere, is to scrap all of this and just take a screenshot of a notebook with this example (and probably a MathJax printing and plotting example as well).

asmeurer:
The build from Travis looks good. I'd like confirmation from @certik (or someone else) that the build on the Docker container also looks good.

certik:
I am testing it now.

certik:
@asmeurer docker works, but you need to apply the following patch:

```
diff --git a/Dockerfile b/Dockerfile
index c7c5b91..09697d1 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
         texlive-fonts-recommended texlive-latex-extra texlive-fonts-extra \
         texlive-humanities texlive-science latex-xcolor dvipng \
         texlive-latex-recommended python-pygments \
+        lmodern texlive-xetex \
         make \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
```

asmeurer:
Does it render correctly?

asmeurer:
Feel free to push up fixes.

certik:
Yes it does.

certik:
Ok, I pushed the changes in.

asmeurer:
OK, feel free to merge if you approve of the change.

certik:
I think that the change looks great. Merging.

asmeurer:
Awesome. Hopefully the journal will have no problems with using DejaVu Sans Mono.

certik:
@asmeurer, we'll submit the pdf of the unicode print, not the xelatex source. So we can use any fonts we want.

asmeurer:
@certik the whole document uses DejaVu Sans Mono now (otherwise the example would look inconsistent).

asmeurer:
Compare before

screenshot 2016-05-13 17 10 17

and after

screenshot 2016-05-13 17 10 30

It's a different font now.

certik:
I see. I think it's fine if the unicode example(s) look a bit different, in case the journal uses some custom font for the main document.

PR #133, opened 2016-05-14
Doctest

isuruf:


printers.tex
```diff
@@ -5,9 +5,11 @@
 this paper so far.
 
 \begin{verbatim}
+>>> from sympy import *
 >>> phi0 = Symbol('phi0')
 >>> str(Integral(sqrt(phi0), phi0))
-Integral(sqrt(phi0 + 1), x)
+'Integral(sqrt(phi0), phi0)'
```
isuruf, line 10: Is this what was intended?

asmeurer, line 10: Yes

asmeurer:
Thanks for doing this.

asmeurer:
I don't like the extra newlines at the end of each example. Can you revert that.

calculus.tex
```diff
@@ -4,21 +4,25 @@
 
 Derivatives can be computed with the \verb|diff| function.
 \begin{verbatim}
+>>> x = symbols("x")
```
asmeurer: The text at the top of the paper says that all examples will assume `from sympy import *` and `x, y, z = symbols('x y z')`.

matrices.tex
```diff
@@ -39,10 +41,11 @@
 >>> S = MatrixSymbol("S", n, p)
 >>> T = MatrixSymbol("t", m, p)
```
asmeurer: This should be changed to `MatrixSymbol("T", m, p)`.

printers.tex
```diff
@@ -22,19 +24,21 @@
 
 \begin{verbatim}
 >>> pprint(Integral(sqrt(phi0 + 1), phi0), use_unicode=False)
-  /
- |
- |   __________
+  /                    
+ |                     
+ |   __________
```
asmeurer: My editor is going to clear these the next time I edit the file. Can we make the doctester ignore trailing whitespace?

test-paper.py
```diff
@@ -0,0 +1,26 @@
...
+        ]
+
+full_text = ""
+
+output_file = "test-paper.txt"
+
+for file in files:
+    with open(file, 'r') as f:
+        full_text += "\n" + f.read()
+
```
asmeurer, line 88: I suppose the empty lines are to make the doctests work. Perhaps `full_text.replace(r'\end{verbatim}', '')` here would have the same effect.

calculus.tex
isuruf: This was needed because `x` is changed at the middle. `x = 1 / (sin(pi/5)+sin(2*pi/5)+sin(3*pi/5)+sin(4*pi/5))**2`
There's also a `x = Symbol('x', positive=True)` statement.

printers.tex
isuruf: Yes, this was changed before `NORMALIZE_WHITESPACE` option was given. Now, these spaces can be removed

isuruf:
There are some differences in doctests for Python 2.7 and Python 3.5 like how sets are printed, etc. What should we use? Python 2.7 or Python 3.5?

certik:
In SymPy somehow the doctests work in all Python versions. Maybe we can do the same.

calculus.tex
asmeurer: We should use different variable names for those.

asmeurer:
Set printing is normally random, but the doctester sorts the output.

isuruf:
What I meant was the printing difference for `set([1, 2])` which is `{1, 2}` in Python 3 and in Python 2, it is the former.

isuruf:
Using SymPy's doctester now. Ready for review

numerics.tex
```diff
@@ -70,8 +70,6 @@ \subsection{The mpmath library}
 >>> import mpmath
 >>> mpmath.mp.dps = 30    # 30 digits of precision
 >>> mpmath.mpf("0.1") + mpmath.exp(-50)
-mpf('0.100000000000000000000192874984794')
->>> print(_)   # pretty-printed
```
asmeurer, line 74: I don't like this change. This loses some information that was portrayed here. And this _is_ the output:

``` py
>>> import mpmath
>>> mpmath.mp.dps = 30    # 30 digits of precision
>>> mpmath.mpf("0.1") + mpmath.exp(-50)
mpf('0.100000000000000000000192874984794')
>>> print(mpmath.mpf("0.1") + mpmath.exp(-50))
0.100000000000000000000192874985
```

series.tex
```diff
@@ -51,6 +51,7 @@ \subsubsection{Series Expansion}
 \texttt{sympy.polys.rs\_series} takes as input any SymPy expression and hence
 there is no need to explicitly create a polynomial \texttt{ring}. An example:
 
+%no-doctest
```
asmeurer: I removed this and got

```
File "/Users/aaronmeurer/Documents/Python/sympy/sympy-paper/test_full_paper.py", line 291, in test_full_paper
Failed example:
    rs_series(sin(a + b), a, 4)
Expected:
    -1/2*(sin(b))*a**2 + (sin(b)) - 1/6*(cos(b))*a**3 + (cos(b))*a
Got:
    -1/6*(cos(b))*a**3 + (cos(b))*a - 1/2*(sin(b))*a**2 + (sin(b))
```

I guess there's a bug here. Sometimes it shows one output and sometimes it shows the other. Did you open an issue for this?

isuruf: Doctests are skipped for `rs_series` in sympy documentation, so I assumed this is a known issue

numerics.tex
isuruf, line 74: I couldn't find another way to get this done. Shall I skip this block?

solvers.tex
```diff
@@ -60,7 +60,7 @@
 >>> solveset(x - x, x, domain=S.Reals)
 (-oo, oo)
 >>> solveset(x - x, x, domain=S.Complexes)
-S.Complexes
```
asmeurer, line 63: This gives `S.Complexes` for me in 1.0.

asmeurer, line 63: Here is the issue:

```
>>> repr(S.Complexes)
'S.Complexes'
>>> str(S.Complexes)
'S.Complexes'
>>> sstr(S.Complexes)
'Complexes((-oo, oo) x (-oo, oo), False)'
```

The doctester should use `repr` (that's what a Python session will use).

asmeurer, line 63: I see. It's used for hash stability. Let's just skip this test (and the mpmath one above).

isuruf, line 63: I think why sstr was used was because of differences in Python 2 and 3 of repr.
Python3

```
In [9]: str(set([(8, 4, 4, 4, 0, 0)]))
Out[9]: '{(8, 4, 4, 4, 0, 0)}'

In [10]: sstr(set([(8, 4, 4, 4, 0, 0)]))
Out[10]: 'set([(8, 4, 4, 4, 0, 0)])'

In [11]: repr(set([(8, 4, 4, 4, 0, 0)]))
Out[11]: '{(8, 4, 4, 4, 0, 0)}'
```

Python2

```
In [5]: str(set([(8, 4, 4, 4, 0, 0)]))
Out[5]: 'set([(8, 4, 4, 4, 0, 0)])'

In [6]: sstr(set([(8, 4, 4, 4, 0, 0)]))
Out[6]: 'set([(8, 4, 4, 4, 0, 0)])'

In [7]: repr(set([(8, 4, 4, 4, 0, 0)]))
Out[7]: 'set([(8, 4, 4, 4, 0, 0)])'
```

asmeurer:
I pushed some fixes, and I'm happy with this now. The tests run for me in both Python 2 and 3. I think the only difference would be the skipped dict.keys test in gruntz.tex, which we use Python 3 format for.

solvers.tex
asmeurer, line 63: I see. I think we ought to use Python 3 printing as our default, now that we don't support Python 2.6 and we can use set literals. I opened https://github.com/sympy/sympy/pull/11116 for this.

asmeurer:
Going to merge this, so that we can avoid merge conflicts and broken doctests for any new changes that come in.

Issue #65, opened 2016-04-19
Funding section


moorepants:
I think we should list all the funding sources for SymPy (grants, etc). This issue can be used to gather a list. It is important to show how grant money supports this kind of work.
- X # of GSoC grants
- This NSF grant funded work on the mechanics package: http://www.nsf.gov/awardsearch/showAward?AWD_ID=0928339

aterrel:
We should also put a reference to NumFOCUS in this section.

asmeurer:
The SIAM template has a specific place for this information.

PR #135, opened 2016-05-16
Added stubbed version of acknowledgements section. 

rpmuller:
Other authors should add their funding agency acknowledgements here.


certik:
Thanks @rpmuller.

hargup:
Should we add GSoC in this section?

moorepants:
see issue #65

PR #137, opened 2016-05-16
Fix Makefile (update authors correctly)

certik:
Previously `make` would not rebuild the `authors.tex` file if you modified `authors.json`.


sumith1896:
+1

asmeurer:
Oops. It's been so long since I've used make.

certik:
You must be using Python or something.

Issue #138, opened 2016-05-16
more interesting integration examples


certik:
we also need more interesting integration examples. can you think of a good example for the meijer g algorithm?

asmeurer:
We should have one indefinite example that uses Risch, and one definite example that uses meijerg.

asmeurer:
And we need a summation example (or two, if we want one for Gosper and one for Meijer G), and a product example (`product` can't do much, so probably just something simple there).

asmeurer:
This is done. Hopefully the examples are good. For meijerg, it took some looking to find an example that didn't return a huge piecewise expression. I ended up doing the Laplace Transform of log(t). For Risch, I just made up an example, which looks complicated enough (but is easy for the algorithm to compute).

PR #140, opened 2016-05-16
Clean up the calculus section a bit

asmeurer:
- Move the integrals to the top
- Focus on the algorithms for each function
- Remove the discussion of unevaluated functions, and just put a single
  sentence about it at the end (with an example)
- Use a slightly more complicated example for derivatives
- Describe the "algorithm" for derivatives
- Consistency in the paragraphs (first sentences names the function,
  subsequent sentences describe the algorithms, then example)

### DIFF
--- a/calculus.tex
+++ b/calculus.tex
@@ -1,4 +1,4 @@
-Integrals are calculated either with the \verb|integrate| function. SymPy
+Integrals are calculated with the \verb|integrate| function. SymPy
 implements a combination of the Risch
 algorithm~\cite{bronstein2005integration}, table lookups, a reimplementation
 of Manuel Bronstein's ``Poor Man's Integrator''~\cite{Bronstein2005pmint}, and
--- a/calculus.tex
+++ b/calculus.tex
@@ -10,7 +10,7 @@ SymPy to compute a wide variety of indefinite and definite integrals.
 >>> integrate(sin(x), x)
 -cos(x)
 \end{verbatim}
-Definite integrals are calculated with the same method, by specifying a
+Definite integrals are calculated with the same function by specifying a
 range of the integration variable. The following computes $\int_0^1\sin(x)\,dx$.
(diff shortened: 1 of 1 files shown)


asmeurer:
Still TODO: we need better integration examples (#138), and summation and product examples.

calculus.tex
```diff
@@ -1,63 +1,40 @@
...
-Unevaluated expressions can be evaluated with the \verb|doit| method.
-\begin{verbatim}
->>> expr.doit()
-cos(x)
-\end{verbatim}
-
-% TODO: A more interesting example here
-Integrals can be analogously, calculated either with the \verb|integrate|
-function, or the unevaluated \verb|Integral| objects.
+Integrals are calculated either with the \verb|integrate| function. SymPy
```
certik: ...or with the `\verb{Integral}` function?

calculus.tex
```diff
@@ -71,3 +48,11 @@
 >>> limit((2*E**((1-cos(x))/sin(x))-1)**(sinh(x)/atan(x)**2), x, 0)
 E
 \end{verbatim}
```
certik, line 50: Reference the section in the supplement here for more details about Limits.

calculus.tex
asmeurer: I've removed all discussion of unevaluated objects from the descriptions here. There is just the paragraph about it at the bottom of this section.

certik:
Otherwise +1

calculus.tex
asmeurer: Oh I see. I forgot to remove "either"

certik: Yes, remove "either".

calculus.tex
certik, line 50: Maybe we shouldn't reference it here, but rather in the introduction.

certik, line 50: I made #141 for this.

PR #143, opened 2016-05-16
[DO NOT MERGE] This should break Travis

certik:
I want to test that Travis breaks if the sympy output is incorrect.

(Close this PR if Travis fails at this change.)


asmeurer:
It works.

certik:
It works, perfect.

PR #144, opened 2016-05-17
Reviews (everyone push to this branch)

asmeurer:
So from now to May 22 (Sunday), we should do the first round of reviews. The goal here is to make the paper coherent. 

**Everyone who wants to be an author needs to do this.**

What you should do now is read through the paper and make any changes you see fit, and push them to this branch (`reviews`). If you do not have push access, let me know and I can add you.  

Everyone pushing to the same branch will minimize merge conflicts. You should constantly be pulling in changes to avoid conflicts, and commit and push as soon as you make a change.

If you don't have time to do a full review, partial reviews are still helpful, and you can also look at the TODO.md file at the root of this repo to see some things that need to be fixed.


asmeurer:
To clarify, this is both for small changes (grammar fixes, adding citations, etc.), and larger changes (rearranging paragraphs/sections, rewriting sections if necessary). Basically, anything that you see as necessary to make the paper high quality and coherent.

introduction.tex
```diff
@@ -1,11 +1,11 @@
 SymPy is a full featured computer algebra system (CAS) written in the Python
 programming language.
 % cite Python?
-It is open source, being licensed under the extremely permissive 3-clause BSD
+It is a free and open source software, being licensed under the 3-clause BSD
 license.
 % cite BSD?
 SymPy was started by Ond\v{r}ej \v{C}ert\'{\i}k in 2005, and it has since
-grown into a large open source project, with over 500 contributors. SymPy is
+grown into a large project with over 500 contributors. SymPy is
```
hargup: Mentioning open source here was redundant.

asmeurer:
Any suggestions on how to fix the non-wrapping line here (top of page 4)?

screenshot 2016-05-17 16 30 28

Is it OK to hyphenate the texttt text? That doesn't even look [so easy to do](https://tex.stackexchange.com/questions/44361/how-to-automatically-hyphenate-within-texttt).

asmeurer:
Nevermind, I changed it to "some of the" and it wraps now :)

certik:
Yeah, it probably had a few words on the line, most of them long and unbreakable, so it overflowed, rather than leave large spaces between words. There is also a command in tex/latex, that allows the paragraph to break, but have large spaces between words, if there is no other option.

Upabjojr:
Will do it tomorrow morning, when you in America are asleep.

asmeurer:
Should the features table contain complete sentences or not? For instance, we currently have

> Calculus: Algorithms for computing derivatives, integrals, and limits.
> Code Generation: Enables generation of compilable and executable code in a variety of different programming languages directly from expressions. Target languages include C, Fortran, Julia, JavaScript, Mathematica, Matlab and Octave, Python, and Theano.

which is inconsistent. We should either have

> Calculus: Contains algorithms for computing derivatives, integrals, and limits.
> Code Generation: Enables generation of compilable and executable code in a variety of different programming languages directly from expressions. Target languages include C, Fortran, Julia, JavaScript, Mathematica, Matlab and Octave, Python, and Theano.

or

> Calculus: Algorithms for computing derivatives, integrals, and limits.
> Code Generation: Generation of compilable and executable code in a variety of different programming languages directly from expressions. Target languages include C, Fortran, Julia, JavaScript, Mathematica, Matlab and Octave, Python, and Theano.

(note the differences at the beginning of the descriptions)

certik:
I merged here the latest master, which now includes the authors.

asmeurer:
@mattpap I pushed some small changes to the polys section here. It was nothing major (just small grammar fixes). The content looks good. However, there are a few places that need citations. Can you add citations for those (I've added comments), as well as a citation for your master's thesis. I've already imported your bib file into our bib file, so the citations you need should already be there.

asmeurer:
@mrocklin the extra spaces around the examples are bad. This makes LaTeX think that the text after each example starts a new paragraph, like

screen shot 2016-05-18 at 2 49 04 pm

mrocklin:
My apologies.  I see that you've resolved the issue.

sumith1896:
Added GSoC to acknowledgements. Can improve with better wording and the actual number of slots SymPy has received over the years.

certik:
The branch is getting pretty long, @asmeurer do you want to merge it into master (but keep the branch intact)? I am hoping that github is clever enough to only start showing commits + diffs from the point where you merge back into master.

asmeurer:
Yeah let's merge it every once in a while. The only way to do it is to merge the PR, meaning we'd have to start a new one each time, but that's fine.

mattcurry:
I've read through the first half of the paper and fixed mainly grammar and flow issues. I've also added a new table entry in the "features" section for quantum mechanics.

fabianp:
I don't know how it is for others but I have compiling errors because of missing `pprint.pdf`

domain_specific.tex
```diff
@@ -4,8 +4,6 @@
 mechanics along with support for manipuating physical quantities with units.
 
 
-\subsection{Classical Mechanics}
```
flacjacket, line 7: This should not have been removed, the below levels are sub-sub-section, this is the relevant sub-section

certik:
Is Travis passing?

asmeurer:
You need to use `make` to build the pdf.

domain_specific.tex
mattcurry, line 7: My bad! I'll fix it on the next PR.

asmeurer:
Going to merge and re-PR as soon as Travis passes. Continue pushing to the same branch.

architecture.tex
```diff
@@ -194,6 +194,8 @@ \subsection{Logical Inference and Assumptions}
 \texttt{y} are both created with \texttt{positive=True}, then
 \verb|(x + y).is_positive| will be \texttt{True}.
 
+%% TODO: axe this? Why mention it if we won't discuss it. Maybe move to
+%% future work. - scopatz
```
asmeurer: I mentioned it because people who know about it would be wondering about it. But putting it in the future work section sound like a good idea. I'll add something now.

matrices.tex
```diff
@@ -25,6 +25,8 @@
 {x - sqrt(y*(x + y)): 1, x + sqrt(y*(x + y)): 1}
 \end{verbatim}
 
+%% TODO a Python list is not a dense representation...
+%% Is this supposed to say flat? - scopatz
```
asmeurer: Yes it is. If you have a 1000 x 1000 matrix, the list representation requires representing all 1000000 elements (even if most of them are zeros).

scopatz: I understand that it is a dense matrix, but the list isn't a dense data structure. A list of lists of symbols would still be a dense matrix but the representation would not be any less dense as a data structure (in Python). I think this could be reworded because it implies that the flatness of the list makes the matrix dense, which is not true.

asmeurer: How is it not a dense data structure?

asmeurer: It isn't really flat either. Does this help? https://github.com/sympy/sympy-paper/pull/144/commits/332203b9ce19aeecf0b703be1f5284c85113b0f9

scopatz: Well a list is collection of `PyObject*`, [namely](http://www.laurentluce.com/posts/python-list-implementation/), 

``` c
typedef struct {
    PyObject_VAR_HEAD
    PyObject **ob_item;
    Py_ssize_t allocated;
} PyListObject;
```

The individual expression elements are all in their own locations on the heap. This is about as non-dense as it gets in terms of memory allocation. 

> It isn't really flat either. Does this help? 332203b

It isn't flat because it is a list of lists? That is fine too. I guess what I am trying to say is that the denseness and the data structure should be independent statements.  Something like, "The Matrix class represents dense matrices. Internally, these matrices store their elements as a list of lists."

asmeurer: You and I are working with different definitions of sparse and dense. The one used here is the same as Jason Grout defines it [here](http://ask.sagemath.org/question/9464/what-is-the-difference-between-dense-and-sparse-matrices/), namely, dense means all values are stored, and sparse means only nonzero values are stored.  Perhaps we should use the specific terminology [LIL](https://en.wikipedia.org/wiki/Sparse_matrix#List_of_lists_.28LIL.29). 

The same usage is used in the polynomials section, by the way,

>  SymPy implements dense and sparse polynomial representations. Both are used in the univariate and multivariate cases. The dense representation is the default for univariate polynomials. For multivariate polynomials, the choice of representation is based on the application. The most common case for the sparse representation is algorithms for computing Gröbner bases (Buchberger, F4, and F5). This is because different monomial orderings can be expressed easily in this representation. However, algorithms for computing multivariate GCDs or factorizations, at least those currently implemented in SymPy, are better expressed when the representation is dense. The dense multivariate representation is specifically a recursively dense representation, where polynomials in K[x0, x1, . . . , xn] are viewed as a polynomials in K[x0][x1] . . . [xn]. Note that despite this, the coefficient domain K, can be a multivariate polynomial domain as well. The dense recursive representation in Python gets inefficient when the number of variables gets high.

scopatz: I don't think we are using different definitions when it comes to the matrices themselves. We both agree that a dense matrix stores all elements.  That said, I agree that we should be as specific as possible.  The LIL that you cited didn't mention a particular language. I'll add that Python is the weird case here where lists are not densely or contiguously stored.  

The polynomial usage is fine because it doesn't tie the densely causally to a Python data structure, which can be said to be non-dense. 

I think the matrix case could be a place where a reviewer would find this objectionable. It is better to be non-ambiguous and prevent such objections.

asmeurer: It sounds like you're conflating dense and contiguous. Python lists may not be contiguous, but they are dense. `[0]*1000` takes (amortized) 1000 as much space to store as `[0]` (same with `[None]*1000`). Javascript arrays (from my understanding of them) are an example of a non-dense array. You can have a Javascript array where the first 1000 items are `undefined` and the 1001st item is `1` and it takes up as much space as `[1]`. The closest thing to a non-dense array in Python is a dictionary indexed by non-negative integers (except you'd need a wrapper because all the methods like len and iteration are wrong).  

If I Google "dense data structure" I only get results that are related to matrices or polynomials. If I Google "dense array" I get references to the non-denseness of Javascript arrays.  Based on that I'd say this is a standard usage of the term. 

I'm all for adding clarity. Should we define "dense" and "sparse"?

scopatz: I can find a number of places where people refer to dense memory allocation when they refer to contiguous chunks:
[here](https://books.google.com/books?id=DqX-aWwIEJsC&pg=PA411&lpg=PA411&dq=dense+memory+allocation&source=bl&ots=a4r5UlBW9u&sig=46mXpruUj7pY4F5_HxoaR9dzvAI&hl=en&sa=X&sqi=2&ved=0ahUKEwj2q8O7sufMAhXDSiYKHf0cBx8Q6AEIKzAD#v=onepage&q=dense%20memory%20allocation&f=false) and [here](https://wiki.dlang.org/Dense_multidimensional_arrays#Dense_dynamic_arrays) seem relevant to the matrix discussion.  I don't think I am the only one who may conflates these as I have heard it used as a term of art, though personally, mostly in database discussions.

Also, I do think that the Python array is a dense/contiguous data structure. Also, if I am not mistaken, slots, strings, bytes, and structs are also all dense/contiguous in their storage. 

Defining the terms and then using them clearly is all that I desired out of my original comment.

scopatz: Also, here is another document that uses the term dense w.r.t. to contiguous memory https://www.kernel.org/doc/ols/2007/ols2007v1-pages-141-152.pdf

asmeurer: I added a footnote to the polys section explaining what they mean (polys because the polys section comes before the matrix section). Hopefully it is sufficient to also clarify the term for the subsequent matrices section, even though the strict definition is slightly different (a matrix has entries and shape; a polynomial has coefficients and degree).

scopatz: I like the footnote, but I would have to read the whole paper through again to see if that section is still in my L1 or L2 by the time I get to this section.  Since the "makes it" wording is still in there, I think that a comment tying this back would make this more explicit, "dense (in the same sense as a "dense polynomial")"., or make that a footnote.

scopatz: Also, this is something that I can do on my next read through

fredrik-johansson:
I've edited my own contributions a bit; please review the changes I made.

I've skimmed most of the rest of the paper, and it looks pretty good. Not sure if I will have time to read everything very carefully.

asmeurer:
Travis finally caught, so I merged this. Continued at https://github.com/sympy/sympy-paper/pull/158.

PR #149, opened 2016-05-18
Make authors style comply with SIAM guidelines

certik:
The guidelines are at:

https://www.siam.org/journals/pdf/stylemanual.pdf
https://www.siam.org/journals/tex/standard/docsiamart.pdf

@leosartaj, @hargup, @shivamvats, @Sumith1896 I had to change the position of the ZIP code in your address, so that the country is at the end, per SIAM guidelines. Can you verify that the ZIP code position is correct, or fix it?


certik:
@asmeurer, this unifies the addresses, as you asked some time ago.

asmeurer:
The style guide also says 

>  Each name is followed by a footnote symbol...repeat each symbol doubled for more than six names.

certik:
@asmeurer I already did this in e1bbe5d392411e21344d7c7a5818055caea02ca7.

certik:
I thought we could also triple and quadruple each symbol, but for now I just double each symbol, per instructions, and then use numbers and let the SIAM editors decide what they want to do.

asmeurer:
I see. It also says you should use the same footnote for duplicate addresses (Harsh and Shivam are the same, Anthony and I will be the same).

certik:
@asmeurer yes, but that's a bit pain to implement automatically. So I will wait with that if the paper gets accepted, then we can easily do this by hand. If the paper is rejected, then other journals might have different styles.

asmeurer:
I agree we should let the journal worry about most of these details. I'm +1 to merge this.

PR #160, opened 2016-05-25
Update patches counts using a newer sympy commit

certik:
Update all relevant files.


scopatz:
Thanks @certik! I'd like to remain last author, please I am not sure if this PR changes that.

certik:
It does change it. We'll have to create a different ordering than by the number of patches. I can do that in a future PR.

scopatz:
Sure, or we could just manually move me to the end at the last minute :)

certik:
@scopatz well, I don't want manual fiddling with authors when we have that many, since that's a recipe for a mistake. So I fixed it in #162.

PR #162, opened 2016-05-25
Introduce an optional author_order

certik:
If present, it will be used instead of sympy_commits to order authors. This
allows us to modify the order of authors. The `author_order` is only used for
ordering, it is not used for anything else (like listing the number of
patches).

Fixes @scopatz's request at https://github.com/sympy/sympy-paper/pull/160#issuecomment-221721955

### DIFF
--- a/authors/authors.json
+++ b/authors/authors.json
@@ -162,15 +162,6 @@
         "sympy_commits": 17
     },
     {
-        "email": "scopatz@cec.sc.edu",
-        "github_id": "scopatz",
-        "institution": "University of South Carolina",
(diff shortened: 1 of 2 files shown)


ashutoshsaboo:
How about if we order the authors in a descending order based on the commits in sympy-paper's repo? Wouldn't that be more relevant since those are like contributions to the paper? I am just asking though. @certik @asmeurer How about if we do it that way?

certik:
I feel it's more relevant to order by commits to sympy itself, and then manually adjust, as we did for Aaron and Anthony.

PR #167, opened 2016-05-27
Remove references to the supplement in the main paper

asmeurer:


certik:
Wait, I am not sure I agree with this. What is the motivation?

certik:
I am -1 on this, unless we incorporate important parts (like citations!) from the supplement into the main article. Let me know what the plan is. Are we going with SISC or SICOMP? If with SICOMP, then we need to incorporate stuff from the supplement into the main article. I can help. Let me know.

asmeurer:
I've already submitted the article to PeerJ preprint and SICOMP.

SICOMP doesn't have supplementary materials, only SISC. I did submit the supplement with the PeerJ preprint. 

I personally think the paper is fine as-is, without any material from the supplement, most of which distracts from the main theme of the paper. If anything, I would move back some of the examples, but I also think the examples we have are fine. 

If you do want to modify this, let me know. I can re-upload the paper (the submission is live), and I suspect the editor won't look at it until Monday or Tuesday (Monday is a holiday).

scopatz:
The Supplement can always go on PeerJ as well.

certik:
Ok, I went over the article. I think it's fine for now. Let's see what the editor and referees say. If it gets rejected, then we might want to rewrite a few things and resubmit or submit somewhere else, we'll see.

scolobb:
The article reads well and looks solid to me: it explains the core of SymPy in sufficient detail to make the reader understand how it works, without boring them too much with technicalities.  There are many examples which make reading even easier.

I believe the only reason the article may get a negative review is because of scope mismatch, but I don't think it's going to happen with SICOMP or PeerJ Computer Science.

PR #176, opened 2016-06-05
Fix DOI information

rouckas:
Some DOI fields contained URL instead of DOI. This commit fixes it.


certik:
Indeed, great catch, the DOI url was there twice in the pdf document. Thanks for fixing it!

PR #177, opened 2016-06-05
Style and other improvements of the Supplement

rouckas:
Mainly unification of the coding style and removal of some useless imports.


certik:
I think that works, tests pass, so I am merging it. Thanks!

Issue #16, opened 2016-03-18
sympy.stats


Upabjojr:
We have a very nice submodule (_sympy.stats_).

@mrocklin has already published something related:
- http://people.cs.uchicago.edu/~mrocklin/tempspace/scipy2012-sympystats-paper.pdf
- https://www.computer.org/cms/ComputingNow/homepage/2012/0712/W_CS_SymbolicStatisticswithSymPy.pdf

I'd suggest to put a short description with proper citations.

sumith1896:
Duplicate of https://github.com/sympy/sympy-paper/issues/126

PR #187, opened 2016-06-09
Add section about sympy.stats

rouckas:
It has become a bit longer than I expected and I don't even mention the sampling functionality. But I can shorten it if needed.


asmeurer:
+1

stats.tex
```diff
@@ -0,0 +1,50 @@
+%% Stats
+The \verb|sympy.stats| module provides random variable types and methods for
+determination of statistical properties of expressions involving random
+variables. The variables can be either continuous or discontinuous, which are
```
Upabjojr: continuous or discrete

stats.tex
```diff
@@ -0,0 +1,50 @@
+%% Stats
+The \verb|sympy.stats| module provides random variable types and methods for
+determination of statistical properties of expressions involving random
```
Upabjojr: _statistical properties_ is a bit ambiguous. What _sympy.stats_ currently does, is to generate the symbolic integrals/summations for statistical expressions. No properties are determined, symbolic calculations are performed.

stats.tex
```diff
@@ -0,0 +1,50 @@
...
+further divided into finite and infinite. They are associated
+with probability densities on corresponding domains and internally defined
+in terms of probability spaces. 
+Apart from the possibility of defining the random variables from user supplied
+density distribution, SymPy provides definitions of most common
+distributions, including \texttt{Uniform}, \texttt{Poisson}, \texttt{Normal},
+\texttt{Binomial}, \texttt{Bernoulli}, and many others.
+
+Properties of random expressions can be calculated using, e.g.,
+\texttt{E} and \texttt{variance} to calculate expectation and variance.
```
Upabjojr: `\texttt{expectation} (abbreviated \texttt{E})`

stats.tex
```diff
@@ -0,0 +1,50 @@
...
+variables. The variables can be either continuous or discontinuous, which are
+further divided into finite and infinite. They are associated
+with probability densities on corresponding domains and internally defined
+in terms of probability spaces. 
+Apart from the possibility of defining the random variables from user supplied
+density distribution, SymPy provides definitions of most common
+distributions, including \texttt{Uniform}, \texttt{Poisson}, \texttt{Normal},
+\texttt{Binomial}, \texttt{Bernoulli}, and many others.
+
+Properties of random expressions can be calculated using, e.g.,
```
Upabjojr, line 13: Expectation, variance, and covariance map random expressions to real numbers. I wouldn't call that properties.

stats.tex
```diff
@@ -0,0 +1,50 @@
...
+\texttt{E} and \texttt{variance} to calculate expectation and variance.
+Conditions on random variables can be defined with inequalities, equalities,
+and logical operators and their overall probabilities are obtained using
+\texttt{P}. The features can be illustrated on a model of two dice throws:
```
Upabjojr, line 21: probability( ) abbreviated P( ). I would like to change _P_ in future versions of SymPy to return an unevaluated probability expression.

stats.tex
```diff
@@ -0,0 +1,50 @@
...
+>>> E(v)                                # mean velocity
+2*sqrt(2)*sqrt(kT)/(sqrt(pi)*sqrt(m))
+>>> E(m*v**2/2)                         # mean energy
+3*kT/2
+>>> solve(density(v)(x).diff(x), x)[0]  # most probable velocity
+sqrt(2)*sqrt(kT)/sqrt(m)
+\end{verbatim}
+
+More information on the \texttt{sympy.stats} module can be found in
+\cite{StatsMRocklin}.
```
Upabjojr, line 57: Are you sure it's the right way to render a citation?

Upabjojr, line 57: @mrocklin are there other papers/links concerning _sympy.stats_ ?

stats.tex
```diff
@@ -0,0 +1,50 @@
+%% Stats
+The \verb|sympy.stats| module provides random variable types and methods for
+determination of statistical properties of expressions involving random
+variables. The variables can be either continuous or discontinuous, which are
+further divided into finite and infinite. They are associated
```
Upabjojr: the latter being further divided

stats.tex
```diff
@@ -0,0 +1,50 @@
...
+
+Properties of random expressions can be calculated using, e.g.,
+\texttt{E} and \texttt{variance} to calculate expectation and variance.
+Conditions on random variables can be defined with inequalities, equalities,
+and logical operators and their overall probabilities are obtained using
+\texttt{P}. The features can be illustrated on a model of two dice throws:
+\begin{verbatim}
+>>> from sympy.stats import Die, P, E
+>>> X, Y = Die("X"), Die("Y")
+>>> P(Eq(X, 6) & Eq(Y, 6))
```
Upabjojr, line 25: Add example with `evaluate=False`, so that it is clear how it works.

stats.tex
```diff
@@ -0,0 +1,50 @@
...
+
+Using the facilities of the \texttt{sympy.stats} module, one can, for
+example, calculate 
+%Using the above-mentioned features, it is easy to calculate, for example
+the well known properties of maxwellian velocity distribution
+\begin{verbatim}
+>>> from sympy.stats import Maxwell, density
+>>> kT, m, x = symbols("kT m x", positive=True)
+>>> v = Maxwell("v", sqrt(kT/m))
+>>> E(v)                                # mean velocity
```
Upabjojr, line 45: Example with `evaluate=False` would be nice here.

stats.tex
scolobb, line 57: To me, citing a resource in this way looks pretty natural.

stats.tex
rouckas: Thanks for the catch, I'll fix it.

stats.tex
Upabjojr, line 57: I'm talking about the rendering style.

stats.tex
rouckas: But the integrals/sums are automatically evaluated. From the end user point of view, E(X) returns the expectation, etc... I would like to leave it as it is for the sake of simplicity, If you agree.

stats.tex
rouckas: OK, I'll change it.

stats.tex
rouckas, line 13: I think that a real number assigned to a mathematical object can be called a property. [Wolfram](https://reference.wolfram.com/language/guide/RandomVariables.html) also calls it properties:

> In the Wolfram Language, you can directly compute several dozen properties from symbolic distributions

stats.tex
rouckas, line 21: OK

stats.tex
Upabjojr: Being a technical paper, I suggest to mention some internal stuff. Most of the algorithms in _sympy.stats_ are about generating integrals/summations. They can be shown with `evaluate=False`.

The term _properties_ is too general, it could mean anything. What about _symbolic values_ instead?

stats.tex
rouckas, line 57: I know about the two papers referenced in #16 

You mean saying "in \cite{StatsMRocklin}"? I think it depends on the final format of refences. But I agree that it may need to be changed, perhaps to "in Ref. \cite..." or "in paper by \cite..." in case of author-year format.

stats.tex
rouckas: OK

stats.tex
rouckas, line 25: OK

stats.tex
rouckas, line 21: However, probability is not exposed when importing from sympy.stats:

```
>>> from sympy.stats import probability
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'probability'

```

One would need to import it from `sympy.stats.rv`

stats.tex
scolobb, line 57: @Upabjojr could you please explain what you mean exactly by rendering style?

In the version I'm looking at I see "can be found in [36]".  That's exactly how I saw citations rendered in very many articles and it looks absolutely natural to me (except that I'd probably add a non-breaking space `~` between "in" and `\cite`.

@rouckas I've never seen "in Ref. \cite" and I find it weird (but then YMMV).  "In the paper by \cite" is even weirder to me, because [36] is a paper, and not a list of authors.  "In the paper \cite" is something see just as often as "in \cite".

rouckas, line 57: @scolobb For example in Journal of Chemical Physics, you could write "in Ref. 11" or "by Esaulov et al.^{11}". I see that peerj template uses author-year citations, so we may be able to write something like "For more information see Rocklin and Terrel (2012)" or "information can be found in (Rocklin and Terrel, 2012)"

stats.tex
rouckas: I will add this information about evaluation

stats.tex
rouckas, line 25: Evaluate=False does not work with finite RV for me:

```
>>> E(Die("X"), evaluate=False)
7/2
```

stats.tex
rouckas, line 45: OK

stats.tex
asmeurer, line 57: Maybe he means to use `in~\cite{StatsMRocklin}`. I'm unclear if `~` is supposed to be used when it's not a parenthetical.

scolobb, line 57: @rouckas All right, I see your point.  Indeed, the citation style varies in different domains.  AFAIK, in maths and computer science the form "in [x]" is quite widespread and accepted.  On the other hand, the way citations are typeset is given by the bibliography style (a LaTeX style file), so I'd go with the following strategy: if the target journal has a recommended bibliography style, we take; if not, we go with what we have now to avoid the extra effort of re-typesetting all the citations :-)

@asmeurer I'd recommend using "in~\cite{...}" because it prevents line breaks between the "in" and the citation number, which (I think) is considered ugly.

rouckas:
@Upabjojr what do you think about the updated text?

stats.tex
Upabjojr: `for determination` or `for the determination`? which one is better?

stats.tex
```diff
@@ -0,0 +1,57 @@
+%% Stats
+The \verb|sympy.stats| module provides random variable types and methods for
+determination of statistical properties of expressions involving random
+variables. The variables can be either continuous or discrete, the latter being
```
Upabjojr: _variables, which can be either_

Upabjojr: _the latter ones_

rouckas: OK, I'll change it to:

```
The \verb|sympy.stats| module provides random variable types and methods for
determination of statistical properties of expressions involving random
variables, which can be either continuous or discrete, the latter ones being
further divided into finite and infinite. The variables are associated
with probability densities on corresponding domains and internally defined
in terms of probability spaces. 
```

(changing also "They" to "The variables" in the second sentence, because it wouldn't be clear what "They" refers to)

stats.tex
rouckas: I prefer `for determination` but I'm not a native speaker

asmeurer: I believe "for the determination" is more grammatically correct, but both sound awkward. "for computing" sounds better.

asmeurer:
@rouckas are all the comments addressed?

rouckas:
@asmeurer I think so. @Upabjojr do you agree?

Upabjojr:
I think this PR is OK.

I would like future versions of SymPy to have _E_ and _P_ return unevaluated expressions whereas the evaluated expression would be returned by _expectation_ and _probability_ (as _Integral_, _Derivative_, _Limit_ vs _integrate_, _derive_, _limit_). Anyway, this is an issue for SymPy.

PR #188, opened 2016-06-09
Style of Calculus section

rouckas:
1) I think that the calculus section needs at least some introductory sentence, so I added one.
2) In my opinion it makes more sense to start with the more basic things - limits and derivatives, so I moved them in front of integrals and summations


asmeurer:
+1

PR #190, opened 2016-06-10
Compile with PeerJ template

rouckas:
These commits contain minimal changes to make the article compile within the peerj template from [Overleaf](https://www.overleaf.com/latex/templates/latex-template-for-peerj-journal-and-pre-print-submissions/ptdwfrqxqzbn). There are still some issues:
- author-year citations are disabled as that would probably require some changes to the text
- the template is not suited for typesetting supplementary information. Formatting and referring to the supplement needs to be resolved
- keywords are missing

and probably others

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -1,8 +1,14 @@
-% SIAM Article Template
-\documentclass[review]{siamart0216}
+%% Submissions for peer-review must enable line-numbering 
+%% using the lineno option in the \documentclass command.
+%%
+%% Preprints and camera-ready submissions do not need 
+%% line numbers, and should have this option removed.
+%%
+%% Please note that the line numbering option requires
+%% version 1.1 or newer of the wlpeerj.cls file.
 
-% Apply a fix to the SIAM Article Template
-\input{siamart0216_uppercase_fix}
+\documentclass[fleqn,10pt,lineno]{wlpeerj} % for journal submissions
+% \documentclass[fleqn,10pt]{wlpeerj} % for preprint submissions
 
 \usepackage{lmodern}
 \usepackage[T1]{fontenc}
@@ -45,12 +51,12 @@
 
 \usepackage{listings}
 
+\newcommand\email[1]{\href{mailto:#1}{#1}}
 \title{SymPy: Symbolic Computing in Python}
 
 \input{authors}
 
-\begin{document}
-\maketitle
+\keywords{Keyword1, Keyword2, Keyword3}
 
 \begin{abstract}
   SymPy is an open source computer algebra system written in pure Python. It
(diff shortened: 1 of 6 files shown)


moorepants:
There is no page limit in PeerJ (I think) so we could get rid of supplements and move them back into the paper. Or just call them appendices.

asmeurer:
> There is no page limit in PeerJ (I think) so we could get rid of supplements and move them back into the paper. 

I am -1 to that. That is a lot of work, and I personally think most of what is in the supplement detracts from the main paper anyway. 

The supplement looks fine to me. What's wrong with it?

wlpeerj.cls
```diff
@@ -41,7 +41,8 @@
                 labelsep=period,%
                 justification=raggedright]{caption}
                 
-\RequirePackage{natbib}
+% custom modification - use numbered citations
+\RequirePackage[numbers]{natbib}
```
asmeurer, line 45: Can this be changed in the tex file itself instead of here?

paper.tex
```diff
@@ -45,12 +51,12 @@
...
 \usepackage{listings}
 
+\newcommand\email[1]{\href{mailto:#1}{#1}}
 \title{SymPy: Symbolic Computing in Python}
 
 \input{authors}
 
-\begin{document}
-\maketitle
+\keywords{Keyword1, Keyword2, Keyword3}
```
asmeurer, line 59: The keywords I used for the preprint and SICOMP were "symbolic, Python, computer algebra system"

asmeurer:
@rouckas I've given you push access. If you want to re-create this PR as a branch on sympy/sympy-paper, that would let the @isuruf-bot build it, and let others push to the branch as well.

moorepants:
> The supplement looks fine to me. What's wrong with it? 

I don't think there is anything wrong with it, it just may be that PeerJ doesn't accept supplements because they don't have a page limit. It's always the case that you have to change the paper to fit the different journal's requirements.

Looks like we can just submit a link to a pdf of the supplement:

https://peerj.com/about/author-instructions/cs#supplemental-information

Note too that we need to figure out how to pay for the PeerJ submission. It will be at least $99 per author\* unless someone negotiates a deal.
- if any authors are already PeerJ members they probably won't have to pay.

asmeurer:
I'll talk with Anthony about the payment. I had forgotten that that was an issue.

wlpeerj.cls
rouckas, line 45: Yes, it can be put into \documentclass.... I will change it tomorrow.

certik:
@rouckas also please add yourself into the authors json file and add your affiliation + address.

rouckas:
@asmeurer I have pushed it to this branch
https://github.com/sympy/sympy-paper/commits/peerj-compile
Is that OK?

Issue #195, opened 2016-06-14
Authorship Criteria


certik:
Dear co-authors (@asmeurer, @smichr, @mattpap, @certik, @skirpichev, @mrocklin, @aktech, @scolobb, @moorepants, @leosartaj, @thilinarmtb, @flacjacket, @ellisonbg, @rpmuller, @Upabjojr, @hargup, @shivamvats, @fredrik-johansson, @fabianp, @mattcurry, @aterrel, @rouckas, @ashutoshsaboo, @isuruf, @Sumith1896, @rc, @scopatz), here are the ICMJE [authorship criteria](http://www.icmje.org/recommendations/browse/roles-and-responsibilities/defining-the-role-of-authors-and-contributors.html#two) that we follow:
1. Substantial contributions to the conception or design of the work; or the
   acquisition, analysis, or interpretation of data for the work; AND
2. Drafting the work or revising it critically for important intellectual
   content; AND
3. Final approval of the version to be published; AND
4. Agreement to be accountable for all aspects of the work in ensuring that
   questions related to the accuracy or integrity of any part of the work are
   appropriately investigated and resolved.

I have verified that all of you satisfy 1. and 2. (and the info is easily accessible in [authors.md](https://github.com/sympy/sympy-paper/blob/9fb21e871bcfa5d0ff3c8dfb70749e7473d5b6d3/authors/authors.md) for anyone to browse). The version to be published is 9fb21e871bcfa5d0ff3c8dfb70749e7473d5b6d3, can you all give a final approval of this version (thus satisfying 3.) and also agree to be accountable for all aspects of the work in ensuring that questions related to the accuracy or integrity of any part of the work are appropriately investigated and resolved (thus satisfying 4.).

Just post a comment that you agree. If you do not agree to 3. or 4., let me know.

These people agreed to 3. and 4. below in a comment:
- [x] @asmeurer
- [x] @smichr
- [x] @mattpap
- [x] @certik
- [x] @skirpichev
- [x] @mrocklin
- [x] @aktech
- [x] @scolobb
- [x] @moorepants
- [x] @leosartaj
- [x] @thilinarmtb
- [x] @flacjacket
- [x] @ellisonbg
- [x] @rpmuller
- [x] @Upabjojr
- [x] @hargup
- [x] @shivamvats
- [x] @fredrik-johansson
- [x] @fabianp
- [x] @mattcurry
- [x] @aterrel
- [x] @rouckas
- [x] @ashutoshsaboo
[...]

PR #196, opened 2016-06-14
Fix rendering of Unicode pprint example

asmeurer:


certik:
That's just a formatting fix, so no need to update #195. (The journal might reformat things further anyway.)

certik:
+1.

PR #198, opened 2016-06-15
Note the SymPy and mpmath versions used for the paper

asmeurer:
One small fix before submission. I noticed that we never mentioned the versions of SymPy and mpmath used for the examples, which is important for reproducibility.


sumith1896:
+1

certik:
+1

PR #210, opened 2016-10-05
Final fixes

asmeurer:
Some final fixes before submission. I plan to submit off this branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,6 +162,7 @@ TSWLatexianTemp*
 authors.tex
 test_full_paper.py
 authors/authors_table.md
+*-diff.tex
 
 # Temp files
 *~
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
 - conda config --set always_yes yes --set changeps1 no
 - conda update -q conda
 - conda info -a
-- conda create -q -n test-environment python=2.7 sympy=1.0 numpy matplotlib
+- conda create -q -n test-environment sympy=1.0 numpy matplotlib xonsh
 - source activate test-environment
 
 # for matplotlib
@@ -23,6 +23,7 @@ before_install:
 script:
 - make
 - make test
+- make diff
 - sh upload_travis.sh
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 before_install:
 - sudo apt-get update && sudo apt-get install -y --no-install-recommends texlive-fonts-recommended
   texlive-latex-extra texlive-fonts-extra texlive-humanities texlive-science latex-xcolor
-  dvipng texlive-latex-recommended python-pygments lmodern texlive-xetex
+  dvipng texlive-latex-recommended python-pygments lmodern texlive-xetex latexdiff
 
 # install sympy, numpy and matplotlib for doctesting
 - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ before_install:
   dvipng texlive-latex-recommended python-pygments lmodern texlive-xetex
 
 # install sympy, numpy and matplotlib for doctesting
-- wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh;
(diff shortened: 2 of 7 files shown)


asmeurer:
So it turns out we need a "track changes" version of the pdf. I am working starting from a xonsh script from @scopatz which uses latexdiff. But there are some issues that I need to work out.

asmeurer:
OK, someone should look over the diff files and make sure they look fine. You can build them with 

```
make diff
```

The diff script requires `xonsh`. Also `latexdiff` (which should come with your TeX installation). This will build `paper-diff.pdf` and `supplement-diff.pdf`. 

Two known issues:
- [ ] Deleted sections (from the supplement) just show up as deleting the section header, because it doesn't pull in the deleted tex files. This can likely be fixed by pulling in the files from the old git tag.
- [ ] The references for the supplement are not diffed. If you uncomment the line in `diff.xsh`, you get the build error:

```
[12 <./images/supp-fig2-integral_steps.png>]) (./comparison_with_mma-diff.tex
[13]) (./supplement-diff.bbl

! LaTeX Error: Something's wrong--perhaps a missing \item.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...

l.29 ...tem[Brent and Zimmermann, 2010]{zimmerman}

!  ==> Fatal error occurred, no output PDF file produced!
```

certik:
Thanks for https://github.com/sympy/sympy-paper/pull/210/commits/2b45728a869f62cd5c43251956913f5391718426, that should do it.

certik:
Here are the current [paper-diff.pdf](https://github.com/isuruf-bot/sympy-paper/raw/16b89d3f41c3af37d362da549b4eb6c319bd3b4a/paper-diff-210.pdf) and [paper-diff-supplement.pdf](https://github.com/isuruf-bot/sympy-paper/raw/16b89d3f41c3af37d362da549b4eb6c319bd3b4a/paper-diff--supplement.pdf).

I am reviewing it now.

certik:
@asmeurer I looked at both diffs, I think they look good. However, I though the table with "projects using sympy" was moved into the main article, but the diff doesn't seem to show it as a new thing.

The minor things you mentioned are fine I think. My understanding is that this is mainly for the editor and reviewers, so that they can check what was changed. If in doubt, they can always consult the old pdf and the new pdf.

asmeurer:
Feel free to push any changes to this branch.

asmeurer:
So I guess the issue I noted above works both ways. Because of the way the includes work, files that were moved to or from the supplement only show a diff for their header. Perhaps we can fix it by diffing them against empty files.

asmeurer:
I don't plan on working on this until tomorrow. If you want to take a stab at it go ahead.

certik:
@asmeurer I am going to bed myself. I think it's fine to leave as is, I would just put this explanation somewhere as a comment/memo when you upload. But if you feel it should be fixed, we can work on in tomorrow.

asmeurer:
Showing blue underlines for the new files is easy enough. We just need to diff against an empty file. 

For the deleted files, it's a little harder. We also can diff against an empty file, but we also would have to add a fake "include" against the empty file in the main document. This isn't a big deal for the supplement, but for the main paper I don't want to upload an empty include file to PeerJ.

asmeurer:
OK, the new sections basic usage and projects that depend on SymPy now show up in blue. I don't want to mess with the deleted ones. The section titles appear. I'll just leave a note to the reviewers that deleted sections only show their section titles in red. 

Now let me see if I can fix the supplement references.

asmeurer:
Fixed the supplement bbl issue. Should be ready to go.

asmeurer:
@certik do you want to give this one last look before I submit?

certik:
Listing 1. on page 19 of paper-diff-210.pdf has some LaTeX garbage in it.

certik:
Otherwise I think it looks good.

asmeurer:
OK, fixed.

asmeurer:
OK, it's been resubmitted! This is the text I put in the note to the staff/editor/reviewers:

> Tables 1, 2, 3 and Listing 1 are in the source file manuscript and can be found in the .tex files.
> 
> The file pprint.pdf has been used to work around issues with displaying Unicode characters in LaTeX. Its contents should appear inline with the text, not as a figure. 
> 
> Due to difficulties with the latexdiff program, in the "track changes" documents, for entire sections that have been removed, only the section titles are shown in red (for example, the "Acknowledgements" section). The removed section content is not shown. Additionally, changes to code examples are not shown.

certik:
Thanks @asmeurer.

I checked and indeed 0c38101 fixed the issue in the diff.

Issue #220, opened 2016-12-08
Proofing PDF


asmeurer:
PeerJ has a PDF for proofing. They want it back within one business day, so if you want to look at it, please do so by tomorrow. 

Here is the PDF: [peerj-cs-103-proof.pdf](https://github.com/sympy/sympy-paper/files/640591/peerj-cs-103-proof.pdf). **Do not circulate this proofing PDF as it will change before publication.**

Here are their instructions:

> **Please do:**
> 1. Clearly mark-up all changes on the proofing PDF downloaded below.
> 2. Read your article carefully and annotate it using Adobe Reader, Preview, or an alternative application that supports PDF annotation. If necessary, submit comments in a text or Word document, referencing line numbers and any relevant query numbers.
> 3. Collect all changes in a single file, whether in the PDF, or a text document.

> **Please do not:**
> 1. Change any of the content: text, figures, tables, or data.
> 2. Edit the original text of the PDF file.
> 3. All changes must be clearly marked-up on the proofing PDF downloaded below. Please do not change any of the text or data, and do not edit the original PDF's content.

> **Check the following**
> - [ ] Are there any numbered queries in the margins (e.g. Q1, Q2)? If so, please address them.
> - [ ] Is the information on page 1 correct? This is where most errors often occur.
> - [ ] Are the figures sharp and clear? If not, please upload identical hi-res figures (minimum 900px x 900px).
> - [ ] Are all the figure and table legend texts complete and accurate?
> - [ ] Is the Additional Information and Declarations section (at the end) correct, and does it read smoothly?

asmeurer:
Please leave any comments about it here and I will include mark them up in the PDF.

certik:
I read through it, and I didn't find any issues. I think it looks great.

asmeurer:
There are a few questions at the end that we need to address.

asmeurer:
I have found and left comments on the following so far:

- The examples have been rendered with curly quotes. They should use straight ASCII quotes.
- The mpmath.mpf example (and likely others) have two single opening quotes and one double closing quote. It should be `mpmath.mpf("0.1")`.
- The Unicode pprint example is rendered too small compared to the surrounding examples. 

I'll upload the final annotated PDF here before I submit it so that people can review the comments.

asmeurer:
I also never got @aktech to confirm the spelling of his name.

asmeurer:
Comments on questions at the end of the PDF:

[...]

PR #221, opened 2016-12-08
Final fixes

asmeurer:
Some final fixes, related to https://github.com/sympy/sympy-paper/issues/220.

As far as I know, we will not be resubmitting the paper itself, only pointing out changes in the annotations on the proof. So changes to the paper here are only for posterity, so that the paper here matches the final published version (at least in content). If someone wants, they can also bring in other changes from the proof (mainly, I noticed they changed the spelling section references in the text, and abbreviated US state names in the affiliations). 

We way wish to resubmit the supplement, as it has removed line numbers, and section numbers starting with S (the PeerJ staff have changed the references in the paper to use the "Section S1" format in their proof).

### DIFF
--- a/authors/authors.json
+++ b/authors/authors.json
@@ -82,7 +82,7 @@
             "Department of Applied Mathematics, Delhi Technological University, New Delhi, India"
         ],
         "institution_address_siam": "Shahbad Daulatpur, Bawana Road, New Delhi 110042, India",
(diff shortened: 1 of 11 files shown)


asmeurer:
If anyone knows how to fix the references (see #220) please feel free to push changes up here, and I will add annotations to the proof (or if you don't want to push a change, just comment here what it should be).

PR #225, opened 2020-05-20
Correct link to galgebra

eric-wieser:
The link here is dead.

xref https://github.com/pygae/galgebra/issues/28

I have no idea whether it makes sense to accept corrections here, but I thought I may as well file the PR.

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -3785,7 +3785,7 @@ Year         = {2016},
   Key          = {galgebra},
   Author       = {Alan Bromborsky},
   Title        = {geometric algebra/calculus modules for SymPy},
-  note         = {\url{https://github.com/brombo/galgebra}},
+  note         = {\url{https://github.com/pygae/galgebra}},
   Year         = {2016},
 }
 
--- a/projects_that_depend_on_sympy.tex
+++ b/projects_that_depend_on_sympy.tex
@@ -42,7 +42,7 @@ Table~\ref{projects-table}.
 \href{http://www.pydy.org/}{\textbf{PyDy}} & Multibody Dynamics with
   Python~\cite{gede2013constrained}. \\
 
-\href{https://github.com/brombo/galgebra}{\textbf{galgebra}} &
+\href{https://github.com/pygae/galgebra}{\textbf{galgebra}} &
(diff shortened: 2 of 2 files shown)


asmeurer:
I don't know if it does either. The published paper already includes the old link. I don't know how this sort of thing is usually handled.

certik:
@brombo, can you please fork the https://github.com/pygae/galgebra repository into your user namespace? Just click on the "Fork" button when logged in as @brombo. That should fix this issue.

certik:
Wait, what happened to https://github.com/brombo ?

certik:
According to this:

https://groups.google.com/d/topic/sympy/2k7105j3Fi0/discussion

Alan's account is now https://github.com/abrombo

asmeurer:
Adding a fork isn't a perfect solution, as the fork would not redirect to the actual repo, where the issues and PRs are. Maybe you can ask GitHub to retroactively make a redirect from brombo/galgebra. They already do this for repos that are moved. 

Otherwise, perhaps one can create the brombo org and put a dummy repo up with a README that points to the new location.

certik:
@eric-wieser would you mind doing what @asmeurer suggested? I think anything along those lines is a way to go.

eric-wieser:
I hadn't thought of recreating the account.

eric-wieser:
> Wait, what happened to https://github.com/brombo ?

He asked GitHub to delete his account, along with the repo and open issues, as he forgot the password and the email address was dead. GitHub refused to reverse it even at his request, since apparently that would need the password or email, even though deletion did not.

asmeurer:
Maybe contact GitHub and see if they can make the redirect. That would be the ideal solution.

certik:
@eric-wieser interesting, I didn't know about this back story. Yes, redirect `brombo` to `abrombo` would be the ideal solution.

eric-wieser:
I went ahead and made the org page, that seemed easier than asking github to create the redirect: 
https://github.com/brombo/galgebra

eric-wieser:
> They already do this for repos that are moved.

I'm pretty sure this becomes invalidated once someone creates a new user with the same name. The redirects are there for convenience, they're not permalinks.

certik:
Perfect! Now I think the problem is fixed.

